PR #7, opened 2025-08-13
feat: handle execution requests and update validator set

matthias-wright:
This PR extends the finalizer to process execution requests. Pending execution requests are stored on disk. So far only deposit requests are processed. Withdrawal request processing will be handled in another PR.

Changes:
- Introduces the [`ExecutionRequest`](https://github.com/SeismicSystems/summit/blob/e6e403120bfe0420c3ce1b7ee5e66f83414125dd/types/src/execution_request.rs#L8) type and some (de)serialization methods
- Implements a [`PersistentQueue`](https://github.com/SeismicSystems/summit/blob/e6e403120bfe0420c3ce1b7ee5e66f83414125dd/utils/src/persistent_queue.rs#L11) type on top of commonware's [`Metadata`](https://docs.rs/commonware-storage/latest/commonware_storage/metadata/struct.Metadata.html) key value storage. This is used to temporarily store execution requests before they are processed
- Extends the deposit request type by adding the ed25519 public key
- Processes the deposit requests by adding the corresponding ed25519 public key to the [registry](https://github.com/SeismicSystems/summit/blob/e6e403120bfe0420c3ce1b7ee5e66f83414125dd/application/src/finalizer.rs#L220) and storing the balance on [disk](https://github.com/SeismicSystems/summit/blob/e6e403120bfe0420c3ce1b7ee5e66f83414125dd/application/src/finalizer.rs#L224) (together with the `withdraw_credentials` etc)
- Switches the p2p network from [lookup](https://docs.rs/commonware-p2p/latest/commonware_p2p/authenticated/lookup/index.html) to [discovery](https://docs.rs/commonware-p2p/latest/commonware_p2p/authenticated/discovery/index.html) and use the genesis validators as [bootstrappers](https://github.com/SeismicSystems/summit/blob/e6e403120bfe0420c3ce1b7ee5e66f83414125dd/node/src/args.rs#L203) 
- Persists the last index in the finalizer to disk

TODOs:
- Process the withdrawal requests
- Handle rewards
- Update registry to handle index properly
- Find way to trait-out processing of deposit requests for Ethereum compatability

### DIFF
--- a/application/src/finalizer.rs
+++ b/application/src/finalizer.rs
@@ -3,6 +3,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use crate::engine_client::EngineClient;
 use alloy_rpc_types_engine::ForkchoiceState;
 use commonware_consensus::Reporter;
 use commonware_macros::select;
@@ -17,11 +18,12 @@ use futures::{
 use metrics::{counter, histogram};
 use rand::Rng;
 use summit_types::Block;
+use summit_types::execution_request::{DepositRequest, ExecutionRequest, WithdrawalRequest};
+use summit_utils::persistent_queue::{Config as PersistentQueueConfig, PersistentQueue};
 use tracing::{info, warn};
-use summit_types::execution_request::ExecutionRequest;
-use crate::engine_client::EngineClient;
 
 const LATEST_KEY: [u8; 1] = [0u8];
+const DEPOSIT_REQUESTS_KEY: [u8; 1] = [0u8];
 
 pub struct Finalizer<
     R: Storage + Metrics + Clock + Spawner + governor::clock::Clock + Rng,
@@ -41,7 +43,13 @@ pub struct Finalizer<
 
     rx_finalizer_mailbox: mpsc::Receiver<(Block, oneshot::Sender<()>)>,
 
-    execution_requests: Vec<ExecutionRequest>,
+    deposit_queue: PersistentQueue<R, DepositRequest>,
+
+    withdrawal_queue: PersistentQueue<R, WithdrawalRequest>,
+
+    validator_onboarding_interval: u64,
+
+    validator_onboarding_limit_per_block: usize,
 }
 
 impl<R: Storage + Metrics + Clock + Spawner + governor::clock::Clock + Rng, C: EngineClient>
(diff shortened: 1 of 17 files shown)


application/src/finalizer.rs
```diff
@@ -141,10 +192,61 @@ impl<R: Storage + Metrics + Clock + Spawner + governor::clock::Clock + Rng, C: E
...
+                                        }
+                                    }
+                                    Err(e) => {
+                                        warn!("Failed to parse execution request: {}", e);
+                                    }
+                                }
+                            }
+
+                            // Add validators that deposited to the validator set
+                            // TODO(matthias): I think `last_indexed` isn't necessary incremented by 1
```
daltoncoder, line 215: It is always incremented by 1 so we are good here

daltoncoder (commented):
Looks good to me! Just a couple things and thoughts

- I think the Cargo.lock is a bit messed up, on a fresh clone i have to rebuild it to get it to compile because it cant find new exports in our types. Can you fresh pull and see if you have same issue
- a couple unused imports in clippy if we can clean up they might be from previous PRs

- Seems we  just automatically add the validator if they deposit anything. We should probably check that the deposit event puts their balance above some threshold. 
How does ethereum handle this? If its enforced at the deposit contract then it is fine how it is but i dont think it is. If ethereum requires a hardfork to update this staking threshold then we can follow and have it as a const or a config somewhere. We can merge this in and tackle this on a future PR.

matthias-wright:
- Oh yeah you're right, the deposit contract only enforces that the deposit is at least 1 eth. I added a minimum stake constant that is enforced in the finalizer.
- I just did a fresh clone and built without any issues. Which rustc version are you on? I'm on `rustc 1.88.0 (6b00bc388 2025-06-23)`

daltoncoder:
> * Oh yeah you're right, the deposit contract only enforces that the deposit is at least 1 eth. I added a minimum stake constant that is enforced in the finalizer.
> 
>     * I just did a fresh clone and built without any issues. Which rustc version are you on? I'm on `rustc 1.88.0 (6b00bc388 2025-06-23)`

Yeah actually i think cargo was caching something it shouldnt have locally for me. i just did a fresh clone again and no isues. Was also using 1.88.0

PR #22, opened 2025-08-19
feat: handle withdrawal requests

matthias-wright:
This adds handling of EIP-7002 withdrawal requests to the application.

Core changes:
- Adds a new [DB](https://github.com/SeismicSystems/summit/blob/m/handle-withdrawal-requests/application/src/db.rs) that contains the entire finalizer state (state variables like `last_indexed`, the pending deposits queue, the pending withdrawals queue, validator accounts). This ensures that the state is consistent, even after shutdowns etc.
- [Adds withdrawal requests to a queue](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L203), after verification. Incoming withdrawal requests are validated as follows:
    - [The validator has to exist on the network](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L205)
    - [The validators balance (minus any pending withdrawals that we already processed) minus the withdrawal amount has to be larger than 0](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L207)
    - [The account that sent the withdrawal transaction has to match the withdrawal credentials](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L209)
    - [If the current withdrawal request will result in the validators balance to go below the minimum stake amount, then the entire validator balance will be withdrawn (this is how validators can exit the network)](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L212)
- Introduces a validator [withdrawal period](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/node/src/engine.rs#L48) . If a withdrawal request was made at height n, then it will be eligible for processing [at epoch n + withdrawal period](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L220)   
- Includes some of the pending withdrawals in the [proposed block](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/actor.rs#L297)
- [Decrement validator balance when we commit a block](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/finalizer.rs#L297) (that included withdrawal requests)
- When a validator verifiies a block, it will also verify the included withdrawal requests. This is done by requesting the next set of withdrawal requests from its own queue, and making sure that they [match the withdrawals from the block.](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/application/src/actor.rs#L336)

TODOs:
- Validator rewards
- Validator status changes. There is already a [type](https://github.com/SeismicSystems/summit/blob/f7024cd32d0978c28dc4ddc8bbe29ab36018cad3/types/src/account.rs#L7) that indicates the status, but it is not being updated currently. 
- Extensive testing (will require some changes to the testing harness, specifically the mock engine client)


application/src/actor.rs
```diff
@@ -255,7 +271,15 @@ impl<R: Storage + Metrics + Clock + Spawner + governor::clock::Clock + Rng, C: E
...
-        let payload_envelope = self.engine_client.get_payload(payload_id).await;
+        let mut payload_envelope = self.engine_client.get_payload(payload_id).await;
+
+        // Add pending withdrawals to the block
+        let withdrawals = pending_withdrawals.into_iter().map(|w| w.inner).collect();
+        payload_envelope
+            .envelope_inner
+            .execution_payload
+            .payload_inner
+            .withdrawals = withdrawals;
```
daltoncoder, line 282: Instead of getting the payload from the engine API and then mutating it to add the withdraws. We need to submit the withdraws as payload attributes when we ask the engine API to build us a block in start_building_block. I suspect not doing it this way will make the state root not match up in new_payload

daltoncoder (changes requested):
Overall everything looks really good. The one change i requested i think is important but i think its quick

PR #29, opened 2025-08-26
feat: use simplex views for registry and update index handling in registry

matthias-wright:
We mixed up the block height and the view

There are three places where the view is incremented in threshold simplex:
- [handle_notarization](https://github.com/commonwarexyz/monorepo/blob/ce5341d4754ebac5e7708111e419fd51c4909660/consensus/src/threshold_simplex/actors/voter/actor.rs#L1175)
- [handle_finalization](https://github.com/commonwarexyz/monorepo/blob/ce5341d4754ebac5e7708111e419fd51c4909660/consensus/src/threshold_simplex/actors/voter/actor.rs#L1318)
- [handle_nullification](https://github.com/commonwarexyz/monorepo/blob/ce5341d4754ebac5e7708111e419fd51c4909660/consensus/src/threshold_simplex/actors/voter/actor.rs#L1232)

The happy path is handle_notarization -> handle_finalization, which will increment the view twice. However, the block height will only be incremented once, when the block is finalized.
Also, in case of a nullification, there is no block added, but the view is still incremented.

This causes the view to outpace the block height. This isn't an issue in itself, however, we've been using the block height to update registry. But simplex will use the view when it calls the registry methods (via the Supervisor traits).

This can cause the following issue. For example, lets say we are at block height 4 and view 6, with 10 validators:
- A new [round](https://github.com/commonwarexyz/monorepo/blob/e3fffd347e68f947a292d322b9fbbd4346428d65/consensus/src/threshold_simplex/actors/batcher/actor.rs#L69) is created for view 6.
- It will call the `participants(6)` method with view 6, which will return a list of the 10 validators.
- A validator is added at height 5.
- From now on, any call to `participants(6)`  will return a list of 11 validators.
- This breaks the contract imposed by the Supervisor trait. For a given view, the returned list of validators has to always be the same.

Changes:
- Adds the view (from simplex) to the [Block](https://github.com/SeismicSystems/summit/blob/819c5c6520371c08f5a240fe2a5b07efaa429745/types/src/block.rs#L27) because the finalizer needs access to it.
- Uses the view instead of the block height for the registry
- Updates the registry to always find the largest view that is <= the requested view

Possible TODOs:
- The [view](https://github.com/SeismicSystems/summit/blob/819c5c6520371c08f5a240fe2a5b07efaa429745/types/src/block.rs#L112) method of the Block returns the height. We should probably change that to the view as well.

### DIFF
--- a/application/src/finalizer.rs
+++ b/application/src/finalizer.rs
@@ -239,6 +239,7 @@ impl<R: Storage + Metrics + Clock + Spawner + governor::clock::Clock + Rng, C: E
                                                         withdrawal_request.amount = remaining_balance;
                                                         account.status = ValidatorStatus::SubmittedExitRequest;
                                                     }
+
                                                     account.pending_withdrawal_amount += withdrawal_request.amount;
                                                     self.state.set_account(&withdrawal_request.validator_pubkey, account).await;
                                                     self.state.push_withdrawal_request(withdrawal_request.clone(), new_height + self.validator_withdrawal_period).await;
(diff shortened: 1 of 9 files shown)


application/src/finalizer.rs
```diff
@@ -340,11 +344,23 @@ impl<R: Storage + Metrics + Clock + Spawner + governor::clock::Clock + Rng, C: E
                                         account.balance = account.balance.saturating_sub(withdrawal.amount);
                                         account.pending_withdrawal_amount = account.pending_withdrawal_amount.saturating_sub(withdrawal.amount);
 
+                                        #[cfg(debug_assertions)]
```
daltoncoder, line 347: Why debug_assertions over "prom" feature flag we were using before? We might want to access these metrics in an optimized build right? Or is this just a metric we only use in tests? In that case maybe we just explicitly use the `test` flag

daltoncoder (approved):
Looks good to me good catch

PR #41, opened 2025-09-15
bring back syncer

daltoncoder:
Bringing back syncer so we can move forward with ed25519 keys until we can use BLS12381 aggregate signatures with simplex


types/src/account.rs
```diff
@@ -45,48 +42,40 @@ impl TryFrom<&[u8]> for ValidatorAccount {
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
         // ValidatorAccount data is exactly 77 bytes
-        // Format: ed25519_pubkey(32) + withdrawal_credentials(20) + balance(8) + pending_withdrawal_amount(8) + status(1) + last_deposit_index(8) = 77 bytes
+        // Format: withdrawal_credentials(20) + balance(8) + pending_withdrawal_amount(8) + status(1) + last_deposit_index(8) = 77 bytes
 
-        if bytes.len() != 77 {
+        if bytes.len() != 45 {
             return Err("ValidatorAccount must be exactly 77 bytes");
```
matthias-wright, line 48: not important, but we should also rename the error message to "ValidatorAccount must be exactly 45 bytes". I should've made this a constant to avoid having to change the value in several places.

PR #75, opened 2025-12-04
fix: use parent_hash comparison instead of block_number for genesis check

crStiv:
Replace block_number == 0 check with parent_hash comparison to correctly
identify genesis block when processing starts from non-zero block numbers.
This ensures proper parent_digest selection regardless of start_block value.


daltoncoder:
Thanks for trying to contribute but your change would be incorrect here
```
                   if parent_digest == genesis_digest {
                        genesis_digest
                    } else {
                        parent_digest
                    }
```
Actually just amounts to always returning parent_digest no matter what and the genesis digest is handled in a special way which is why we have the original conditional

PR #92, opened 2026-01-13
add parent block to checkpoint

daltoncoder:
Adds the parent block to the checkpoints. This is needed because you need information from the parent block to propose. 

Changed --checkpoint-or-default to a bool to indicate if you want to start node normally with no checkpoint

### DIFF
--- a/types/src/checkpoint.rs
+++ b/types/src/checkpoint.rs
@@ -59,7 +59,7 @@ impl SszEncode for Checkpoint {
         let data_vec: Vec<u8> = self.data.as_ref().to_vec();
 
         data_vec.ssz_bytes_len()
-            + ssz::BYTES_PER_LENGTH_OFFSET * 2  // 1 variable-length field needs 1 offset
+            + ssz::BYTES_PER_LENGTH_OFFSET * 2  // 2 variable-length field needs 2 offset
             + 32 // digest as [u8; 32]
             + self.parent.ssz_bytes_len()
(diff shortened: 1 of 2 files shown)


types/src/block.rs
```diff
@@ -247,7 +247,7 @@ impl ssz::Decode for Block {
 
 impl EncodeSize for Block {
     fn encode_size(&self) -> usize {
-        self.ssz_bytes_len() + ssz::BYTES_PER_LENGTH_OFFSET
```
matthias-wright: why not use this const?

types/src/checkpoint.rs
```diff
@@ -44,15 +50,18 @@ impl SszEncode for Checkpoint {
...
+        encoder.append(&self.parent);
         encoder.finalize();
     }
 
     fn ssz_bytes_len(&self) -> usize {
         let data_vec: Vec<u8> = self.data.as_ref().to_vec();
 
         data_vec.ssz_bytes_len()
-            + ssz::BYTES_PER_LENGTH_OFFSET  // 1 variable-length field needs 1 offset
+            + ssz::BYTES_PER_LENGTH_OFFSET * 2  // 1 variable-length field needs 1 offset
```
matthias-wright: not important, but the comment still says 1 field instead of 2

daltoncoder: fixed

PR #112, opened 2026-02-05
perf: avoid duplicate block broadcast in application actor

VolodymyrBg:
Before: The application actor cached the last built block and rebroadcast it on Relay::broadcast, causing a full Block clone and a second syncer.proposed(...) call for the same proposal.
Now: The application actor sends the proposed block to the syncer only once in the Propose path, treats Broadcast as a no-op, and removes the unused built_block cache and Arc<Mutex> dependency.


daltoncoder:
its a good catch calling syncer.proposed() twice was not intended. We do have a bit more involved refactor of this section underway in-house though so im going to close for now

PR #162, opened 2026-04-16
feat: add disk metrics to prom

matthias-wright:
Adds reporting for total bytes, used bytes, and available bytes, for each mountpoint.


matthias-wright:
Accidentally merged https://github.com/SeismicSystems/summit/pull/163 before this one, so this commit was merged with https://github.com/SeismicSystems/summit/pull/163

PR #167, opened 2026-05-05
feat: check/execute EL payload on certify

matthias-wright:
Currently the EL payload is not verified by Simplex. This is an optimization to avoid EL payload execution on the consensus hot path. Instead, the EL payload of notarized/finalized blocks is verified when the blocks are executed by the finalizer actor.
If a block contains an invalid EL payload, Reth will reject and skip it. The block stays in Summit's chain in case it was a finalized block. Summit keeps track of Reth's forkchoice.
The consequence of this is that Summit's height might be larger than Reth's height.

This PR explores an alternative approach, where the EL payload is verified in `CertifiableAutomaton::certify`. This will prevent finalizations of blocks with invalid EL payload.
Benchmarks show that this will reduce the blocks per second by around 5.9%.

Changes:
- Call `check_payload` in `CertifiableAutomaton::certify`
- Discard notarized blocks in the finalizer actor if the EL payload is invalid (no chance of finalization)
- Throw an error and initiate shutdown in the finalizer actor if a finalized block contains an invalid EL payload

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6051,6 +6051,7 @@ dependencies = [
  "alloy-eips",
  "alloy-primitives",
  "alloy-rpc-types-engine",
+ "anyhow",
  "bytes",
  "commonware-codec",
  "commonware-consensus",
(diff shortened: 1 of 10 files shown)


evonide:
LGTM.

FYI: Tempo validates the EL payload during `verify` and leaves Commonware `certify` as default-true whereas we instead use `certify` as a heavier pre-finality EL validity gate which is a reasonable fix here.

Recommendation: add a small unhappy-path test showing `certify` returns false when `check_payload` rejects the payload.

PR #169, opened 2026-05-12
fix: verify prev_epoch_header_hash

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/175.

Changes:
- Verifies that the `prev_epoch_header_hash` field on a proposed block matches the expected value. 
- Removes the unused block_value field from the block header.


evonide:
This fixes live boundary-header linkage, but issue #175 is about checkpoint import accepting an unlinked finalized-header chain. Please add the same adjacency check to checkpoint verification: each imported finalized header should link to the previous one and the chain should end at the checkpoint being verified.

matthias-wright:
Done!

evonide:
LGTM! This now seems to correctly reject live boundary blocks with the wrong previous-header hash and checkpoint imports whose finalized headers do not form one hash-linked chain.

PR #170, opened 2026-05-13
fix: verify proposed block's round and refactor header digest calculation

matthias-wright:
This addresses https://github.com/SeismicSystems/summit/issues/173.

Edit: this also addresses: https://github.com/SeismicSystems/summit/issues/196

Changes:
- Adds checks to `handle_verify` to ensure that the Simplex view and epoch matches the view and epoch from the proposed block's header.
- Implements `ssz_derive::Encode` and `ssz_derive::Decode` for `Header` and `Block` . This avoids scenarios where a field is added the structs, but not included in the digest calculation. The `Header` now has to be created before we can call `encode` to get the serialized bytes, so we cannot directly have a `digest` field on the header anymore. A `get_digest` method is added, and a field `digest: OnceLock<Digest>` is added to the header to avoid re-calculating the digest whenever `get_digest` is called.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6199,6 +6199,7 @@ dependencies = [
  "dirs 6.0.0",
  "ethereum_hashing",
  "ethereum_ssz",
+ "ethereum_ssz_derive",
  "futures",
  "libc",
  "metrics",
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -790,19 +790,20 @@ fn handle_verify<ES: Epocher>(
         );
         return false;
     }
-    if block.header.prev_epoch_header_hash != aux_data.header_hash {
+    if block.header.prev_epoch_header_hash() != aux_data.header_hash {
         warn!(
             "prev_epoch_header_hash mismatch: expected: {:?}, received: {:?}",
-            aux_data.header_hash, block.header.prev_epoch_header_hash
+            aux_data.header_hash,
+            block.header.prev_epoch_header_hash()
         );
         return false;
     }
 
     // Validate consensus trie state root
-    if block.header.parent_beacon_block_root != aux_data.state_root {
+    if block.header.parent_beacon_block_root() != aux_data.state_root {
         warn!(
             expected = ?aux_data.state_root,
-            actual = ?block.header.parent_beacon_block_root,
+            actual = ?block.header.parent_beacon_block_root(),
             "parent_beacon_block_root mismatch"
         );
         return false;
@@ -811,30 +812,32 @@ fn handle_verify<ES: Epocher>(
     // Validate checkpoint_hash (None means [0; 32], matching Block::compute_digest)
     let expected_checkpoint_hash: Digest =
         aux_data.checkpoint_hash.unwrap_or_else(|| [0; 32].into());
-    if block.header.checkpoint_hash != expected_checkpoint_hash {
+    if block.header.checkpoint_hash() != expected_checkpoint_hash {
         warn!(
             expected = ?expected_checkpoint_hash,
-            actual = ?block.header.checkpoint_hash,
+            actual = ?block.header.checkpoint_hash(),
             "checkpoint_hash mismatch"
         );
         return false;
     }
 
     // Validate added_validators
-    if block.header.added_validators != aux_data.added_validators {
+    let added_validators = block.header.added_validators();
+    if added_validators != aux_data.added_validators {
         warn!(
             expected_count = aux_data.added_validators.len(),
-            actual_count = block.header.added_validators.len(),
+            actual_count = added_validators.len(),
             "added_validators mismatch"
         );
         return false;
     }
 
     // Validate removed_validators
-    if block.header.removed_validators != aux_data.removed_validators {
+    let removed_validators = block.header.removed_validators();
+    if removed_validators != aux_data.removed_validators {
         warn!(
             expected_count = aux_data.removed_validators.len(),
-            actual_count = block.header.removed_validators.len(),
+            actual_count = removed_validators.len(),
             "removed_validators mismatch"
         );
         return false;
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -749,6 +749,22 @@ fn handle_verify<ES: Epocher>(
         return false;
     }
     // Basic structural validation
+    if block.view() != round.view().get() {
+        warn!(
+            "block view mismatch: expected: {}, received: {}",
+            round.view().get(),
+            block.view()
+        );
+        return false;
+    }
+    if block.epoch() != round.epoch().get() {
+        warn!(
+            "block epoch mismatch: expected: {}, received: {}",
+            round.epoch().get(),
+            block.epoch()
+        );
+        return false;
+    }
     if block.parent() != parent.digest() {
         warn!(
             "block parent mismatch: expected: {}, received: {}",
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -346,6 +346,8 @@ pub fn capture_state_root(&mut self, el_block_number: u64) {
 
 This is called after `execute_block` in the finalizer. The frozen `proof_tree` is used for all subsequent proof generation (via RPC), while the live `ssz_tree` continues to be mutated by finalization operations. The snapshot includes the sorted validator keys (needed for positional index lookups) and the withdrawal pubkey index (stored inside the `SszStateTree`).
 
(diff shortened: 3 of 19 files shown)


evonide:
The live proposal path looks addressed. Could you also confirm the same round/header binding is enforced for sync/imported blocks and same-digest re-proposals and not only normal proposal verification?

matthias-wright:
Regarding the same-digest re-proposals, I'm not sure if this check is necessary, since the block was already verified in the first proposal. and it is confirmed that the digests are matching. We also know that the view won't match, since this is a re-proposed block from a previous view.

Regarding the synced blocks, the syncer checks the following for finalized blocks:
```rust
if block.height() != height
    || finalization.proposal.payload != block.digest()
    || finalization.epoch() != epoch
{
    response.send_lossy(false);
    return false;
}
```
and for notarized blocks:
```rust
if notarization.round() != round || notarization.proposal.payload != block.digest()
{
    response.send_lossy(false);
    return false;
}
```
These checks are the same as in marshal.

However, the difference is that in Summit we specifically added epoch and view to the block (header),so we might have to uphold this invariant in the syncer.

This would mean extending the checks for finalized blocks to something like:
```rust
let certified_round = finalization.round();
if block.height() != height
    || finalization.proposal.payload != block.digest()
    || finalization.epoch() != epoch
    || block.epoch() != certified_round.epoch()
    || block.view() != certified_round.view()
{
    response.send_lossy(false);
    return false;
}
```  

and for notarized blocks:
```rust
let certified_round = notarization.round();
if certified_round != round
    || notarization.proposal.payload != block.digest()
    || block.epoch() != certified_round.epoch()
    || block.view() != certified_round.view()
{
    response.send_lossy(false);
    return false;
}
```
Let me know what you think about this.

evonide:
Ack re same-digest re-proposals makes sense to skip this here as it doesn't make sense given old views can still be valid.

The proposed fix for syncing makes sense please incorporate that and we'll take another look.

evonide:
A critical blocker here: This refactor currently drops `new_with_verify` which is essential to ensure a decoded block’s body actually matches the hashes committed in its header. Without it SSZ decode can produce a block whose payload or execution requests do not match the signed/header commitments.

matthias-wright:
Update (b82fcfd):
- Dropped Decode from #[derive(...)] (kept derived Encode).
- Restored a manual ssz::Decode for Block that decodes the three fields and runs new_with_verify, which re-hashes the body and rejects any block whose payload/execution_requests don't match the header's committed hashes.

evonide:
This now restores `new_with_verify` and fixes the decoded body/hash mismatch blocker from the latest update.

However, there are two remaining concerns.

1) Before merging, this PR looks stale against the final #189 boundary fix. The current `handle_verify` shape puts the same-digest boundary branch before the `aux_data.epoch == round.epoch()` check that was added there. If this is just a stacking/rebase issue then rebasing onto current base and preserving #189’s guard should address it.

2) Same-digest reproposals reuse the old block bytes so the header still contains the original view. If that same digest is certified in a later view, live verification can accept it but sync import now rejects it because it requires the header view to equal the certificate view. Could valid boundary reproposal certificates fail during catch-up unless sync import handles this same-digest exception too? A clean fix may be to keep strict header/round equality for normal blocks and only relax the view check for a certified same-digest reproposal of the final block of the epoch.

Small follow-up: `Header::new_with_digest` is still public and can pair arbitrary header fields with an arbitrary digest. Current usage looks genesis-only but could this be made private/genesis-only or validate the supplied digest to avoid future regressions?

matthias-wright:
Update (6b442fd):
- Rebases this branch onto the base branch. This should solve issue 1.
- Makes `Header::new_with_digest`

Regarding issue 2:
I'm thinking about removing from the `view` field from `Header`.
- This removes the view binding from blocks and solves the re-proposal edge case.
- It also frees us from always having to verify that the view in the finalization certificate matches the view in the header.

The only reason why the `view` was added to `Header` is to pass it to the syncer on a (re-)start: https://github.com/SeismicSystems/summit/blob/6b442fd4e3ee63ed41cfbbb4500b48957c1bcfcd/syncer/src/config.rs#L14
The syncer uses these values to initialize `last_processed_round`: https://github.com/SeismicSystems/summit/blob/6b442fd4e3ee63ed41cfbbb4500b48957c1bcfcd/syncer/src/actor.rs#L421
`last_processed_round` is only used in two places:
1. To guard against stale subscriptions: https://github.com/SeismicSystems/summit/blob/6b442fd4e3ee63ed41cfbbb4500b48957c1bcfcd/syncer/src/actor.rs#L682
2. For cache-pruning: https://github.com/SeismicSystems/summit/blob/6b442fd4e3ee63ed41cfbbb4500b48957c1bcfcd/syncer/src/actor.rs#L1268
Furthermore, `last_processed_round` will be updated on the next finalization: https://github.com/SeismicSystems/summit/blob/6b442fd4e3ee63ed41cfbbb4500b48957c1bcfcd/syncer/src/actor.rs#L1276

So the concrete fix would be:
- remove `view` from `Header`
- remove the checks  that bind the header to a specific view
- pass in view 0 to the syncer (the view resets every epoch to 0)

Let me know what you think!

evonide:
Thanks, I think we are almost there. Removing `view` from the header might work but it feels like a larger protocol change. If you are confident there are no knock-on effects that path is fine.

The smaller fix may be to keep the strict header/certificate view check for regular blocks but align sync/import with live verification for same-digest boundary reproposals. If a later-view certificate is for the exact same epoch-boundary block, sync/import should not reject it just because the reused block header still has the original view.

I checked Tempo/Commonware and their sync/import behavior is consistent with the live reproposal path here.

Could you either remove `view` from the header with tests around restart/sync or keep `view` and add the same-digest boundary reproposal exception in sync/import ideally with a regression test?

matthias-wright:
Update(fa54b9c);
-Allow same-digest epoch-boundary reproposals in sync import

evonide:
I double checked the sync/import side because this invariant is important to get right. I think there is still one concrete gap before this is safe to approve: the same block/finalization mismatch is checked when both arrive together but can be missed when the finalization is cached first and the block arrives later by digest.

The direct finalized/notarized response paths now check the block against the certificate round which is the right idea:
https://github.com/SeismicSystems/summit/blob/fa54b9c950aac8a9bb2c7e2fd366085b6b34e8c9/syncer/src/actor.rs#L939-L1048

The issue is that a finalization can also be cached first and the block can arrive later through the plain block fetch path. That path only checks the block digest before looking up the cached finalization:
https://github.com/SeismicSystems/summit/blob/fa54b9c950aac8a9bb2c7e2fd366085b6b34e8c9/syncer/src/actor.rs#L912-L937

Both delivery orders then reach `store_finalization` but that helper stores the block/finalization pair without enforcing the same block-vs-certificate round check:
https://github.com/SeismicSystems/summit/blob/fa54b9c950aac8a9bb2c7e2fd366085b6b34e8c9/syncer/src/actor.rs#L1392-L1435

So the root cause seems to be that the validation lives on some ingress branches instead of at the join point where a block/finalization pair becomes trusted storage. The lowest-churn fix may be to run the block-vs-finalization validation inside `store_finalization` whenever `finalization` is `Some(_)`. The existing branch-level checks can stay as early rejection, but `store_finalization` should be the final guard so cached-finalization + later block fetch, local cached block + finalization, and gap repair all get the same protection.

That helper should still preserve the same-digest final-block reproposal exception: strict epoch/view binding for normal blocks, and the explicit final-block same-digest exception where the reused block’s header view can be older than the certificate view.

matthias-wright:
Update(14cafc6):
- Added the requested block-vs-certificate round check in store_finalization

This change broke the [test_finalize_same_height_different_views](https://github.com/SeismicSystems/summit/blob/fa54b9c950aac8a9bb2c7e2fd366085b6b34e8c9/syncer/src/lib.rs#L1191) test, which tested that the same block can be finalized under different views by different nodes, and each persists whatever finalization it saw.
The test was inherited from commonware's `marshal` and still exists on the main branch today: https://github.com/commonwarexyz/monorepo/blob/2027158f5c0af24eae4c0595a46390ab0187e8cc/consensus/src/marshal/mocks/harness.rs#L4943

The reason this test failed is because it uses a block at height 1 (with epoch length 20). I fixed the test by using a block at height 19 (last block of an epoch), because with these new changes we only accept a block view that is different from the finalization view on the last block of the epoch.

This makes the syncer stricter than `marshal`. On marshal this test passes on every block, not just the last block of the epoch.

Just for reference: Tempo also add the view to the [block](https://github.com/tempoxyz/tempo/blob/8eabc9420685564748ef24ef1b6f61fff1643e82/crates/primitives/src/header.rs#L19), but uses `marshal` (which doesn't check that block view == finalization view).

evonide:
LGTM! This looks solid now thanks for your diligence :)!

PR #178, opened 2026-05-13
fix: applying protocol params backwards

### LINKED ISSUE #176: [Security] High (correctness): protocol parameter updates are applied in reverse order
## Claim

Protocol-parameter changes are queued in submission order but applied in reverse order at epoch transition. Because updates are appended to a `Vec` and later drained with `pop()`, the last submitted update is applied first and an older update to the same parameter can overwrite the newer intended value.

## Impact

If multiple updates to the same protocol parameter are accepted in one epoch, Summit can end the epoch with the wrong final value. For example, if `MaximumStake(40 ETH)` is submitted before `MaximumStake(64 ETH)`, the expected final value is 64 ETH, but LIFO application can apply 64 first and then overwrite it with 40.

This can leave stale operational parameters active for stake bounds, epoch length, timestamp future limits, deposit caps, withdrawal caps, or observer limits. The bug is especially relevant when chained with attacker-controlled or unauthenticated parameter updates, but the ordering defect exists even if the updates are otherwise authorized.

## Code

- Grouped protocol-parameter requests are decoded in byte order:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/execution_request.rs#L281
- Finalizer preserves parse order when ingesting protocol-parameter requests:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1699
- `push_protocol_param_change` appends each update to the end of a `Vec`:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L343
- Epoch transition applies queued changes by popping from the end of the vector:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L544
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L698
- Affected fields include stake bounds, epoch length, timestamp skew, deposit cap, withdrawal cap, and observers per validator:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L702
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L707
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L712
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L719
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L727
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L731

## Context

Summit allows protocol parameters to be changed through execution requests. These requests are parsed in observed order and queued in consensus state for application at an epoch boundary.

The expected semantics for repeated updates to the same parameter are normally last-write-wins: if two updates to `MaximumStake` are submitted in one epoch, the later update should determine the final value. Summit's queueing logic preserves submission order when pushing updates, but the application logic reverses that order by using `pop()`.

As a result, if two updates affect the same parameter, the older value can become the final live parameter. This is not merely a display or ordering issue; these parameters affect validator admission, stake-bound enforcement, epoch scheduling, timestamp validation, deposit throughput, withdrawal throughput, and observer authorization.

The issue is distinct from whether protocol-parameter requests are authorized. Missing authorization can make it easier for an attacker to submit malicious updates, but even an authorized sequence can be applied with the wrong final semantics because of the LIFO queue drain.

daltoncoder:
Fixes #176

### DIFF
--- a/types/src/consensus_state.rs
+++ b/types/src/consensus_state.rs
@@ -697,7 +697,7 @@ impl ConsensusState {
 
     pub fn apply_protocol_parameter_changes(&mut self) -> bool {
         let mut min_or_max_stake_changed = false;
-        while let Some(param) = self.protocol_param_changes.pop() {
+        for param in self.protocol_param_changes.drain(0..) {
             match param {
                 ProtocolParam::MinimumStake(min_stake) => {
                     self.validator_minimum_stake = min_stake;


sebastian-osec:
Looks good to me.

Issue #186, opened 2026-05-14
[Security] Critical (attackable): Last-Block Joining Validator Withdrawal Leaves Stale Header Addition


sebastian-osec:
## Context

At epoch boundaries, Summit records validator-set transition data in finalized block headers. Those header deltas, including `added_validators` and `removed_validators`, are later replayed by checkpoint and sync verifiers to reconstruct the validator committee history.

For the epoch's last block, Summit builds auxiliary header data from the parent state. That parent is the penultimate block state. The invariant is that the validator-transition data committed in the last-block header must describe the same activation and removal effects that live canonical state applies when entering the next epoch.

## Claim

A joining validator whose activation is scheduled for the next epoch can submit a withdrawal that is included in the epoch's final block. Summit snapshots `added_validators` from the penultimate parent state into the last-block header before processing that same block's execution requests. The withdrawal then cancels the validator's pending activation in live state, but the finalized header still carries the stale `added_validators` entry.

## Impact

Live nodes and finalized-header reconstruction can permanently disagree about whether the validator joined. Live canonical state removes the pending activation and does not activate the validator, while checkpoint and header-chain verifiers replay the finalized header and reconstruct the validator as added.

This can make otherwise honest checkpoint bundles fail trustless verification, prevent new nodes from bootstrapping through the documented checkpoint path, and make later certificates appear to come from a committee that differs from the one live nodes actually used.

## Root Cause

The last-block header transition data is captured from the penultimate parent state before same-block execution requests are applied. The joining-validator withdrawal path can mutate the same `added_validators` transition state after the header has already committed to the old snapshot.

## Code

- Last-block auxiliary data includes `added_validators` for the next epoch in the block header: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L960
- Application proposal builds the block from finalizer auxiliary data before the finalizer later processes that block's execution requests: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L551
- New-validator deposits schedule activation through `added_validators` at a future epoch: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1773
- A withdrawal by a joining validator removes the pending future activation from live state: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1641
[...]

PR #187, opened 2026-05-14
fix: buffer last-block withdrawals for pending validators

matthias-wright:
Note: this PR builds on https://github.com/SeismicSystems/summit/pull/168 because it addresses a similar edge case, and the code changes are overlapping.

This addresses https://github.com/SeismicSystems/summit/issues/186.
Withdrawals that are arriving on the last block of an epoch for joining validators that are awaiting their participation in the validator set have to be postponed until the next epoch.
This avoids an edge case where the finalized header still includes them as an added validator for the next epoch, because it was created on the penultimate block of the epoch.

Changes:
- Buffer withdrawal requests for joining validators that arrive on the last block of an epoch until the next epoch.
- Sets the status for validators whose participation was aborted to `Inactive`.
- Adds an e2e test `test_joining_validator_withdrawal_on_last_block_keeps_header_consistent` that verifies that the finalized header is consistent even when a withdrawal for a joining validator arrives on the last block of an epoch.
- Adds an e2e test `test_joining_validator_withdrawal_inline_cancel_clears_status` that verifies that the status of joining validators whose participation was aborted by a withdrawal request is correctly changed to `Inactive`.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -1728,21 +1728,7 @@ async fn parse_execution_requests<
                                 let remaining_balance = account.balance;
                                 withdrawal_request.amount = remaining_balance;
 
-                                // If the validator is in the warm-up phase after depositing the stake
-                                // and before joining the committee, then the onboarding is aborted
-                                if account.joining_epoch > state.get_epoch() {
-                                    // Cancel validator's pending activation
-                                    if state
-                                        .remove_added_validator(account.joining_epoch, &public_key)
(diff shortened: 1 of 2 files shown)


sebastian-osec:
LGTM. This PR buffers last-block withdrawals for joining validators so the already-captured `added_validators` header delta remains consistent. It also carries the foundational stake-bound delta fix from https://github.com/SeismicSystems/summit/pull/168.

Issue #180, opened 2026-05-13
[Security] High (correctness): stake-bound changes can poison pending deposits


evonide:
## Context

Summit represents a new validator deposit before processing by creating an inactive validator account with zero balance and `has_pending_deposit = true`. Later deposit processing credits the funds and clears the deposit flag.

Stake-bound enforcement is meant to apply to validator balances during epoch processing. It should not treat a temporary zero-balance pending-deposit placeholder as a real underfunded validator that needs withdrawal handling.

## Claim

A normal stake-bound protocol update can run while a new-validator deposit is queued, causing Summit to mark the temporary zero-balance pending-deposit account as `has_pending_withdrawal = true`; after the deposit is later credited, that stale withdrawal flag remains and blocks future top-ups or withdrawals.

## Impact

A valid validator deposit can be credited and activated while carrying a stale pending-withdrawal flag. The validator's funds can then become stuck behind normal request validation because future withdrawal and top-up requests see an already-pending withdrawal.

## Root Cause

Stake-bound enforcement scans pending-deposit placeholder accounts by balance and can enqueue a zero-amount withdrawal for them. Deposit processing clears only `has_pending_deposit`, and the zero-deduction withdrawal path does not clear the stale `has_pending_withdrawal` flag.

## Code

- New-validator deposits create an inactive account before funds are credited: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1502.
- That placeholder starts with `balance: 0`, `has_pending_deposit: true`, and `has_pending_withdrawal: false`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1523.
- Stake-bound enforcement scans validator accounts by balance without excluding inactive pending-deposit placeholders: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1224.
- The below-minimum branch sets `has_pending_withdrawal = true` and queues a withdrawal for the current balance, which is zero for the placeholder: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1242, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1256.
- Deposit processing later clears only `has_pending_deposit`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1739.
[...]

PR #188, opened 2026-05-14
fix: skip validators with pending deposits for stake-bound balance checks

matthias-wright:
Note: this PR builds on https://github.com/SeismicSystems/summit/pull/187 because it addresses a similar edge case, and the code changes are overlapping.

This addresses https://github.com/SeismicSystems/summit/issues/180.

Changes:
- When checking validator balances after changing `MinimumStake`, only validators without pending deposits are considered.
- An e2e test `test_stake_bound_skips_pending_deposit_placeholder` is added that verifies that inactive validator accounts with zero balance and `has_pending_deposit = true` are skipped when enforcing stake bounds
- An e2e test `test_stake_bound_refunds_top_up_when_max_lowered_before_processing` is added that verifies that when a top-up deposit is queued before a max stake reduction makes the post-top-up balance invalid, deposit processing refunds the top-up and leaves the validator's pre-top-up balance and flags untouched.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -1339,6 +1339,14 @@ impl<
 
             for (key, balance, withdrawal_credentials) in validators_to_process {
                 if balance < self.canonical_state.get_minimum_stake() {
+                    // Nothing to withdraw and nothing in the committee to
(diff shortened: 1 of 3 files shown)


sebastian-osec:
Looks good to me.

Issue #183, opened 2026-05-14
[Security] Critical (attackable): Post-boundary child blocks can be certified


evonide:
## Context

Summit validators verify proposed blocks before voting on them. Around epoch boundaries, this verification is especially important because the last block of an epoch must hand control to epoch-transition logic instead of allowing normal same-epoch child construction to continue.

Honest proposers already avoid building a normal child after an epoch-terminal parent. They re-propose the terminal block instead. Validators must enforce the same rule when verifying blocks from other proposers, because a Byzantine proposer is not constrained by the honest proposal path.

## Claim

A Byzantine proposer can submit a normal child block after an epoch-terminal parent, and Summit’s verifier accepts it as a valid ordinary child because it checks parent and height continuity but does not reject non-reproposal children past the epoch boundary.

## Impact

Honest validators can certify a block that should be impossible under the epoch-boundary state machine. That threatens epoch-transition safety because the certified child can bypass the required boundary transition path and create state that does not match the intended validator-set and epoch progression.

In the worst case, nodes may notarize or execute a fork state that skipped required epoch-transition logic, or later fail when that old-epoch child is applied after canonical state has advanced.

## Root Cause

`handle_verify` special-cases same-digest re-proposals, then accepts ordinary children when `block.height() == parent.height() + 1`. It does not also check whether the parent is already the last block of the epoch, nor does it enforce that a non-reproposal child height is still inside the current epoch.

## Code

The bug is the mismatch between the honest proposer path and the verifier path.

Honest proposer path rejects this shape:

- When the parent is already the last block of the epoch, honest proposal does not build a normal child. It re-proposes the terminal block instead: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L465

Verifier path does not enforce the same rule:

- `handle_verify` only special-cases same-digest re-proposals, then continues into ordinary child verification for other blocks: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L605
- The ordinary child check accepts `block.height() == parent.height() + 1`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L622
- There is no corresponding check that rejects a non-reproposal child when the parent is already epoch-terminal: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L598

Downstream paths can then treat the impossible child as real consensus input:

[...]

PR #189, opened 2026-05-14
fix: reject non-reproposal child after epoch terminal parent

matthias-wright:
This addresses https://github.com/SeismicSystems/summit/issues/183.

Changes:
- Adds a check to `handle_verify` to ensure post-boundary blocks have to be re-proposal blocks.
- Adds a unit test `rejects_non_reproposal_child_after_epoch_terminal_parent` that verifies that a non-reproposal child is rejected after the epoch boundary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6026,6 +6026,8 @@ dependencies = [
 name = "summit-application"
 version = "0.0.3-alpha"
 dependencies = [
+ "alloy-primitives",
+ "alloy-rpc-types-engine",
  "anyhow",
  "commonware-consensus",
  "commonware-cryptography",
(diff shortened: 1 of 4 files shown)


evonide:
This fixes the direct post-boundary child case when verifier aux data is still for the parent’s epoch.

One concern is that handle_verify uses aux_data.epoch to decide whether a block is past the epoch boundary but the verify path does not seem to check that this aux-data epoch matches `round.epoch()`. If verification is still processing an old-epoch block while the finalizer has already moved to the next epoch could it use next-epoch aux data and incorrectly accept a block that should be rejected at the boundary? It seems that verification should first require `aux_data.epoch == round.epoch()` before applying this boundary check.

matthias-wright:
Makes sense, I added the check for the epoch.
I also snuck in a comment about the withdrawal amount endianness, which is relevant for https://github.com/SeismicSystems/summit/issues/262.

evonide:
Thanks LGTM!

PR #190, opened 2026-05-14
fix: enforce consistency between CL header metadata and EL payload metadata

### LINKED ISSUE #185: [Security] Critical (attackable): CL Header Metadata Is Not Bound to EL Payload Metadata
## Context

Summit blocks combine consensus-layer header metadata with an embedded execution-layer payload that is later validated through the Engine API. Honest proposal construction normally keeps these two metadata sources aligned: the Summit header height and timestamp describe the same block as the embedded execution payload's block number and timestamp.

That binding must be enforced by validators when they verify received proposals. Proposer-side construction conventions are not a security boundary, especially as proposer participation becomes less permissioned, because a Byzantine eligible proposer can construct block objects that honest code would not produce.

## Claim

A Byzantine proposer can submit a Summit block whose consensus header passes Summit's parent, height, timestamp, state-root, checkpoint, validator-delta, and withdrawal checks while the embedded execution payload carries a different block number or timestamp. Summit does not require `payload.block_number == header.height` or `payload.timestamp == header.timestamp` before the block can be accepted by the application verifier.

## Impact

Summit can certify and finalize headers and checkpoints whose consensus metadata describes one height or time while the embedded execution payload executes under different EL metadata. This breaks the expected one-to-one meaning of a Summit block across CL and EL consumers.

The mismatch can bypass Summit's CL timestamp policy, confuse proof or checkpoint consumers that trust the Summit header, and create CL/EL consistency failures for descendants and recovery flows. If the mismatched payload is EL-valid, nodes can execute under EL metadata that disagrees with the finalized Summit header; if it is EL-invalid, the existing finalizer path can still advance CL height and head after logging the invalid payload.

## Root Cause

Summit binds the payload bytes to the header through `payload_hash`, but it does not enforce semantic consistency between duplicated CL and EL metadata fields. Engine API validation cannot enforce this relationship because Reth validates the EL payload against EL rules and does not know Summit's `header.height` or `header.timestamp`.

## Code

- Summit blocks carry a Summit `Header` and an independent `ExecutionPayloadV3`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L15
- `Block::compute_digest` hashes the EL payload into `payload_hash` while separately passing CL `height` and `timestamp` into the header: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L51, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L71
- Block decoding checks that the payload hash matches the header payload hash, but does not compare EL payload number/timestamp against CL header height/timestamp: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L94
- Honest proposal construction passes a local timestamp into Reth and then builds the Summit header, but this is proposer-side convention rather than verifier enforcement: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L515, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L551
- Remote verification checks Summit parent, height, timestamp, and wall-clock bounds without an Engine API call in `handle_verify`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L598, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L613, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L622, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L638
- Verification continues through state root, checkpoint, validator deltas, and withdrawals without binding EL payload metadata to the Summit header: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L648, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L670, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L690, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L702
- EL payload validation happens later in finalizer execution; if invalid, the finalizer logs the failure but still advances Summit latest height, view, and head digest: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L49, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1331, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1398, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1414
- Finalizer advances CL state from the Summit header while proof metadata is taken from EL payload block metadata: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1414, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1459

## Fix

Enforce the CL/EL metadata binding during application verification before validators vote or notarize a proposal. Reject any block where `block.payload.block_number != block.header.height` or `block.payload.timestamp != block.header.timestamp`, and keep this check alongside any pre-vote Engine API validation. Add regression tests showing that verifier acceptance fails for otherwise valid blocks with mismatched Summit header height/timestamp and EL payload block number/timestamp.

matthias-wright:
Note: this PR builds on https://github.com/SeismicSystems/summit/pull/189 because it addresses a similar edge case, and the code changes are overlapping.

This fixes https://github.com/SeismicSystems/summit/issues/185.

Changes:
- Adds a check to `handle_verify` to enforce that Summit header's block number matches the payload's block number.
- Adds a check to `handle_verify` to enforce that Summit header's timestamp matches the payload's timestamp.
- Adds unit tests that ensure that blocks that violate these constraints are rejected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6026,6 +6026,8 @@ dependencies = [
 name = "summit-application"
 version = "0.0.3-alpha"
 dependencies = [
+ "alloy-primitives",
+ "alloy-rpc-types-engine",
  "anyhow",
  "commonware-consensus",
  "commonware-cryptography",
--- a/application/Cargo.toml
+++ b/application/Cargo.toml
@@ -36,5 +36,5 @@ alloy-rpc-types-engine.workspace = true
 
 [features]
 prom = ["metrics"]
-bench = []
+bench = ["summit-types/bench"]
 permissioned = []
--- a/application/Cargo.toml
+++ b/application/Cargo.toml
@@ -30,6 +30,10 @@ metrics = { version = "0.24.0", optional = true }
 # commonware depends on it, so we have to build it anyway.
 prometheus-client = "0.22.3"
 
+[dev-dependencies]
+alloy-primitives.workspace = true
+alloy-rpc-types-engine.workspace = true
+
 [features]
 prom = ["metrics"]
 bench = []
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -652,6 +652,24 @@ fn handle_verify<ES: Epocher>(
         return false;
     }
 
+    // Bind the embedded EL payload's metadata to the Summit header.
+    let payload_block_number = block.payload.payload_inner.payload_inner.block_number;
+    if payload_block_number != block.height() {
+        warn!(
+            header_height = block.height(),
+            payload_block_number, "payload.block_number does not match header.height"
+        );
+        return false;
+    }
+    let payload_timestamp = block.payload.payload_inner.payload_inner.timestamp;
+    if payload_timestamp != block.timestamp() {
+        warn!(
+            header_timestamp = block.timestamp(),
+            payload_timestamp, "payload.timestamp does not match header.timestamp"
+        );
+        return false;
+    }
+
     // Validate consensus trie state root
     if block.header.parent_beacon_block_root != aux_data.state_root {
         warn!(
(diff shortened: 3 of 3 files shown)


evonide:
LGTM. This PR binds the EL payload block number and timestamp to the Summit header during verification so mismatched CL/EL metadata can no longer pass vote-time checks.

matthias-wright:
Cool! I'm guessing this means that https://github.com/SeismicSystems/summit/pull/189 is approved as well, since this one builds on  https://github.com/SeismicSystems/summit/pull/189?

evonide:
Commented on one pending issue in #189 we should resolve before proceeding here.

sebastian-osec:
The only special case I noticed is the same-digest re-proposal path, which returns early before these new payload metadata checks. That seems acceptable because it only accepts a re-proposal of the already-known epoch-terminal parent block, not a fresh child with attacker-chosen payload metadata. I do not think that blocks this fix.

matthias-wright:
I accidentally merged this PR into main.
I will revert that commit and merge https://github.com/SeismicSystems/summit/pull/429 into audit-may-2026 instead.

Issue #184, opened 2026-05-14
[Security] Critical (attackable): Blob Versioned Hashes Are Dropped From Summit Blocks


sebastian-osec:
## Context

Summit proposals carry execution payload data from the proposer through consensus so peers can later validate and execute the same payload against their execution clients. For Prague/V4 payloads with blob transactions, Engine API validation needs the expected blob versioned hashes derived from the payload envelope's blob commitments.

That blob commitment data is part of the trust boundary between the consensus block and the execution client. The same execution payload that consensus certifies must carry enough committed metadata for every peer to call `engine_newPayloadV4` with the expected versioned hashes and validate the payload consistently.

## Claim

When a Summit proposer builds or includes a blob-bearing V4 execution payload, Summit commits only the execution payload and execution requests into the consensus block. It drops the V4 payload envelope's blob bundle and commitment data needed to derive the expected versioned hashes. Peers later validate the same payload through `engine_newPayloadV4` with an empty versioned-hash list, so any blob-bearing payload is validated against incomplete Engine API inputs. The normal trigger is an otherwise valid blob transaction being included in a proposed execution payload.

## Impact

Blob-bearing execution payloads lose the consensus-committed versioned-hash data required for correct Engine API validation. An otherwise valid blob-bearing payload can be submitted to peers' execution clients with `versioned_hashes = []`, causing the execution layer to reject it even though Summit consensus has already accepted the block representation that dropped the blob metadata.

Summit can then advance consensus state over a block whose embedded execution payload was not faithfully revalidated with the blob commitments required by `engine_newPayloadV4`. This creates a CL/EL consistency failure for blob-bearing blocks and can break descendant execution, finalized-state consistency, and recovery behavior once blob transactions are included.

## Root Cause

Summit's block format, codec, proposal path, and execution-validation path carry only the execution payload and execution requests. They do not preserve the blob bundle, blob commitments, or derived versioned-hash vector needed for V4 payload validation, and `EngineClient::check_payload` therefore calls `new_payload_v4` with `Vec::new()` for versioned hashes.

## Code

- `Block` contains only `header`, `payload`, and `execution_requests`, so there is no committed field for blob bundle, commitments, or versioned hashes: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L15
[...]

PR #191, opened 2026-05-14
fix: reject blocks that contain blob-bearing transactions

matthias-wright:
Note: this PR builds on https://github.com/SeismicSystems/summit/pull/190 because it addresses a similar edge case, and the code changes are overlapping.

This addresses https://github.com/SeismicSystems/summit/issues/184.

We currently don't support blobs, so this PR adds checks that reject blob-bearing transactions.

Changes:
- Adds checks to `handle_verify` that ensure that `blob_gas_used` is 0.
- Adds a unit test to ensure that blocks with `blob_gas_used > 0` are rejected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6026,6 +6026,8 @@ dependencies = [
 name = "summit-application"
 version = "0.0.3-alpha"
 dependencies = [
+ "alloy-primitives",
+ "alloy-rpc-types-engine",
  "anyhow",
  "commonware-consensus",
  "commonware-cryptography",
--- a/application/Cargo.toml
+++ b/application/Cargo.toml
@@ -36,5 +36,5 @@ alloy-rpc-types-engine.workspace = true
 
 [features]
 prom = ["metrics"]
-bench = []
+bench = ["summit-types/bench"]
 permissioned = []
--- a/application/Cargo.toml
+++ b/application/Cargo.toml
@@ -30,6 +30,10 @@ metrics = { version = "0.24.0", optional = true }
 # commonware depends on it, so we have to build it anyway.
 prometheus-client = "0.22.3"
 
+[dev-dependencies]
+alloy-primitives.workspace = true
+alloy-rpc-types-engine.workspace = true
+
 [features]
 prom = ["metrics"]
 bench = []
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -652,6 +652,24 @@ fn handle_verify<ES: Epocher>(
         return false;
     }
 
+    // Bind the embedded EL payload's metadata to the Summit header.
+    let payload_block_number = block.payload.payload_inner.payload_inner.block_number;
+    if payload_block_number != block.height() {
+        warn!(
+            header_height = block.height(),
+            payload_block_number, "payload.block_number does not match header.height"
+        );
+        return false;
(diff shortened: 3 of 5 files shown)


evonide:
LGTM. This PR rejects blob-consuming payloads before consensus accepts them so Summit no longer certifies blob payloads without the versioned-hash data needed for Engine API validation.

sebastian-osec:
I agree this rejects the honest/common blob-consuming payload case where `blob_gas_used > 0`, but I think the security invariant needs to be enforced from transaction contents, not only from the payload’s declared blob gas field.

`blob_gas_used` is part of the proposer-supplied payload. A Byzantine proposer could set `blob_gas_used = 0` while including an EIP-4844/blob transaction in `payload.payload_inner.payload_inner.transactions`. In that case `handle_verify` would pass the new check, while `EngineClient::check_payload` still calls `engine_newPayloadV4(..., Vec::new(), ...)`.

The added test also uses `transactions: Vec::new()` and manually sets `payload.blob_gas_used = 131_072`, so it proves the declared-gas branch rejects, but it does not test an actual blob transaction or the attacker-controlled-field bypass.

I think the PR should either:
1. parse the execution payload transactions and reject any EIP-4844/blob transaction, or
2. carry consensus-committed versioned hashes/blob commitment data through `Block` and pass them to `new_payload_v4`.

matthias-wright:
Yeah, I agree that the `blob_gas_used` guard is not sufficient to prevent blob transaction.
I do think that rejecting certain transaction types should be the responsibility of the EL.
Summit should not inspect EL payloads directly.
Does option 2 imply removing the `blob_gas_used` guard, and letting the EL handle this completely?

sebastian-osec:
I do not think option 2 requires removing the `blob_gas_used` guard. Keeping it as a cheap early rejection for the common unsupported-blob case seems fine.

But I think the actual #184 fix should still be to carry consensus-committed blob validation data through `Block` and pass it to `new_payload_v4`.

The reason is that `blob_gas_used` is only a proposer-supplied payload field, so it is not strong enough to be the security boundary by itself. `new_payload_v4` is the EL validation point, and for V4 blob payloads it needs the expected `versioned_hashes`. Right now Summit still calls `new_payload_v4(..., Vec::new(), ...)`, so the block format still cannot faithfully revalidate blob-bearing payloads.

So I think the guard can stay, but the complete fix is to commit the versioned hashes/blob commitment data in the Summit block and thread that into `EngineClient::check_payload`.

matthias-wright:
I agree that `blob_gas_used` can be easily altered by the proposer.

However, since all honest validators will pass `versioned_hashes=[]` to `new_payload_v4`, the EL will reject any payloads that contain blob transactions: https://github.com/SeismicSystems/seismic-reth/blob/4bbc6502ea566171decc97adc0c39106967e5ad9/crates/payload/validator/src/cancun.rs#L93

All honest validators call `new_payload_v4` in `CertifiableAutomaton::certify`, preventing a blob-bearing payload from getting finalized.

sebastian-osec:
Got it. I agree that if Summit's current invariant is "no blob transactions are supported", then passing versioned_hashes=[] to new_payload_v4 makes the EL reject actual blob transactions even if blob_gas_used is spoofed to 0.

Issue #181, opened 2026-05-13
[Security] High (correctness): Engine API errors panic live consensus paths


evonide:
## Context

Summit wraps Reth Engine API calls for block building, payload retrieval, payload validation, and forkchoice updates. These calls are used on live proposal, finalization, and fork-tracking paths, so callers need recoverable results for invalid payloads, syncing execution clients, transport failures, and normal JSON-RPC rejections.

The intended boundary is that Reth errors should not abort the validator process. Summit should return a typed error or retryable status so consensus code can reject, defer, or shut down in a controlled way.

## Claim

A validator whose Engine API call receives a non-transport JSON-RPC error can panic inside Summit's Engine API wrapper instead of returning a typed recoverable result. This can happen on local proposal, payload-validation, or forkchoice paths.

## Impact

A stale local payload ID, EL configuration mismatch, invalid forkchoice state, or transient Reth JSON-RPC rejection can crash the validator instead of producing a controlled retry or invalid result. If the same finalized or notarized state can trigger the same Reth error on multiple nodes, validators can enter a correlated crash loop.

## Root Cause

The Engine API client treats non-transport RPC errors as impossible and uses `expect`/panic on live consensus paths instead of returning typed errors to callers.

## Code

- Block-building forkchoice update panics on non-transport errors: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L122.
- Payload retrieval panics on non-transport errors: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L145.
- `new_payload_v4` checking panics on non-transport errors: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L173.
- Forkchoice commit panics on non-transport errors: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L191.
- Proposals call `start_building_block` and then `get_payload`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L515, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L540.
- Finalization calls `check_payload`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1331.
[...]

PR #192, opened 2026-05-14
fix: return errors from engine client methods instead of panicing

matthias-wright:
Note: this PR builds on https://github.com/SeismicSystems/summit/pull/167 because it addresses a similar edge case, and the code changes are overlapping.

This addresses https://github.com/SeismicSystems/summit/issues/181.

Changes:
- Refactors engine client to return errors and remove the `expects` calls.
- In the application and finalizer, where engine client methods are called, handle errors with a graceful shutdown (transport errors are already retried inside of the engine client, only non-transport errors are returned).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6180,6 +6180,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-provider",
  "alloy-rpc-types-engine",
+ "alloy-transport",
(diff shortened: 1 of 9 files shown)


evonide:
LGTM! This removes the live Engine API expect panic path and seems to handle returned errors at proposal, certify, and finalizer call sites correctly.

Issue #182, opened 2026-05-14
[Security] High (attackable): Unauthenticated RPC can sign caller-chosen validator deposits


evonide:
## Context

Summit exposes a JSON-RPC service for node/client operations while the node process also has access to local validator signing keys. Deposit signatures are sensitive because they authorize a validator identity together with withdrawal credentials, so any RPC path that asks the node to sign deposit data needs to stay behind an operator/admin trust boundary.

Summit documents `getDepositSignature` as the staking flow because the relevant keys are only available from inside the node environment. That can be a reasonable operational design if the RPC port is strictly firewalled, but the current implementation does not enforce that assumption in code.

## Claim

A caller who can reach an insufficiently firewalled Summit RPC listener can invoke `getDepositSignature` with attacker-chosen amount and withdrawal address parameters, causing the node to sign deposit credentials with its local validator keys and return usable deposit signatures.

## Impact

For an unstaked node, this can let an untrusted network caller preemptively register the node’s validator identity with attacker-chosen withdrawal credentials. For an already-staked validator, the endpoint still exposes a dangerous remote signing oracle over validator keys and enables unauthorized deposit flows tied to the victim’s identity.

This should be treated as a high-severity trust-boundary issue, not as proof that every production node is internet-exposed. Operators may intend to rely on host firewalls, cloud security groups, private subnets, or container networking, but Summit does not make that dependency explicit or enforce method-level protection. As a comparison point, Ethereum validator/key APIs are generally treated as administrative surfaces and are expected to be localhost-only, explicitly enabled, authenticated, or otherwise firewalled.

## Root Cause

`getDepositSignature` performs privileged validator-key signing for caller-supplied deposit data without authentication, operator approval, withdrawal-address policy, or local-only binding guarantees. The sensitive signing operation is exposed through the normal Summit RPC listener.

## Code

- `getDepositSignature` is exposed as a public RPC method: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/api.rs#L44
- The request supplies `amount` and caller-controlled `address`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/api.rs#L47
- The server embeds the caller-controlled address into withdrawal credentials: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L224
- The server reads local consensus and node private keys before signing: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L230
[...]

PR #193, opened 2026-05-15
fix: create admin RPC server for getDepositSignature endpoint

matthias-wright:
This addresses https://github.com/SeismicSystems/summit/issues/182.

Changes:
- Creates a new `SummitAdminApi` trait served on a separate listener bound to `127.0.0.1` and a separate admin port.
- Moves the `getDepositSignature` endpoint to the new trait.


sebastian-osec:
Looks good to me.

Issue #174, opened 2026-05-13
[Security] Critical (attackable): duplicate consensus keys can panic epoch entry


evonide:
## Claim

Summit does not enforce uniqueness of BLS consensus public keys across validator accounts. Because validator accounts are keyed by Ed25519 node public key, an actor who controls an already-used BLS consensus key can register a second validator identity under a fresh node key using the same BLS key. When both identities become active, epoch scheme construction panics on the duplicate BLS key.

## Impact

Nodes that process the affected epoch transition can panic during orchestrator scheme construction, halting block production. The duplicate key is accepted through normal deposit processing because the new node key is fresh and both the node and BLS signatures verify, but the active validator set later contains two node identities mapped to the same BLS consensus key. `BiMap` construction rejects duplicate BLS values and the code panics.

## Code

- Deposit/account lookup is keyed by Ed25519 node public key:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1494
- New validator accounts store the submitted BLS consensus key without checking global uniqueness:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1523
- Deposit verification proves possession of the submitted node key and BLS key, but does not scan existing accounts for duplicate BLS keys:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1921
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1999
- Deposit processing later queues the validator with that same BLS consensus key:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1782
- Active validators are collected as `(node_key, consensus_key)` pairs without BLS-key deduplication:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L652
- Epoch transition sends the active validator set to the orchestrator:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L676
- Scheme construction builds a `BiMap` and panics if duplicate BLS values exist:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/scheme.rs#L109
- Orchestrator constructs the epoch scheme during epoch entry:
  https://github.com/SeismicSystems/summit/blob/ed2c5c8/orchestrator/src/actor.rs#L236

## Context

[...]

PR #194, opened 2026-05-15
fix: enforce BLS key uniqueness

matthias-wright:
This addresses https://github.com/SeismicSystems/summit/issues/174.

Changes:
- Adds checks to ensure that a new joining validator can re-use an existing BLS key.
- Adds an e2e test `test_duplicate_bls_consensus_key_rejected` that verifies that a deposit request with a duplicate BLS key is rejected.
- Adds checks to ensure that a top-up deposit from an existing validator uses the same BLS key as the initial deposit.
- Adds an e2e test `test_top_up_deposit_with_mismatched_bls_key_rejected` that verifies that a top-up deposit with a different BLS key is rejected.


sebastian-osec:
Looks good to me.

Issue #179, opened 2026-05-13
[Security] High (correctness): Engine payload timestamp uses milliseconds


evonide:
## Context

Summit uses wall-clock timestamps in its consensus layer and also asks Reth to build execution payloads through Engine API `PayloadAttributes`. The consensus side tracks time in milliseconds, while the execution layer interprets the payload timestamp as EVM `block.timestamp`, which is Unix seconds.

That CL/EL boundary needs an explicit unit conversion. Otherwise the same numeric timestamp can pass Summit's consensus checks while becoming a completely different time value inside the EVM.

## Claim

Normal block production takes Summit's millisecond wall-clock timestamp and passes it unchanged into the Engine API payload timestamp field, so honest proposers can build execution payloads whose EVM `block.timestamp` is roughly 1000x larger than the intended Unix-seconds value.

## Impact

Honest validators can certify and finalize execution payloads with broken EVM time. Contracts that depend on `block.timestamp`, including timelocks, expiries, vesting schedules, TWAP windows, and other time-based logic, will observe millisecond-scale timestamps instead of normal Unix seconds.

## Root Cause

Summit uses one millisecond timestamp value for both consensus timing checks and Engine API payload construction, even though the Engine API timestamp field is seconds-based.

## Code

- Proposal timestamp is read from `epoch_millis()`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L480.
- The same millisecond `current` value is passed into `start_building_block`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L515.
- `start_building_block` forwards that value unchanged into Engine API `PayloadAttributes { timestamp, ... }`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L101.
- Summit verification also compares block timestamps against millisecond local time, so it does not catch the CL/EL unit mismatch: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L638.
- Genesis documents the timestamp future-drift parameter in milliseconds: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L48.
- Lighthouse computes the expected execution payload timestamp from slot start as seconds and rejects payloads whose `execution_payload.timestamp()` differs: https://github.com/sigp/lighthouse/blob/stable/beacon_node/beacon_chain/src/execution_payload.rs#L318-L331.
[...]

PR #198, opened 2026-05-18
chore: add comment for millis timestamp

matthias-wright:
This addresses https://github.com/SeismicSystems/summit/issues/179.


evonide:
LGTM thanks for adding this!

Issue #197, opened 2026-05-16
[Security] High (correctness): Finalizer Catch-Up Can Deadlock While Local EL Stays SYNCING


evonide:
## Context

Summit's finalizer actor applies notarized and finalized blocks, advances local consensus state, serves auxiliary-data requests, and coordinates epoch progress. It also calls the execution layer through `check_payload` before applying a block, because a local node may need to validate or execute blocks that it did not personally certify.

PRs #167 and #192 appear to mitigate the normal certification path by adding execution-payload checks during `CertifiableAutomaton::certify` and returning Engine client errors instead of panicking. However, the finalizer still needs its own execution-layer check for lagging validators, restarting validators, and checkpoint/catch-up paths where local certify did not run.

## Claim

A lagging, restarting, or catching-up validator whose local execution layer keeps returning `SYNCING` can deadlock its finalizer actor. The finalizer-side `execute_block` path retries `SYNCING` indefinitely inside the actor task, so the node can stop processing unrelated finalizer mailbox work while waiting for local EL recovery.

This is narrower than the original pre-#167 certification issue: after PR #167, ordinary block certification should require a quorum whose execution layers returned `VALID`. The remaining issue is the finalizer fallback/catch-up path, where the local node imports or applies already-certified/finalized blocks without having completed local certify.

## Impact

An affected validator can stop making local finalizer progress during catch-up or recovery. It may stop handling auxiliary-data requests, finalized-block updates, notarized fork work, epoch progress, and cancellation-sensitive finalizer work until the execution layer recovers or the operator intervenes.

This is primarily affected-node liveness and recovery DoS. It can become broader only if the same EL-syncing condition affects many validators at once, but the post-#167 normal certification path no longer makes network-wide finality halt the default impact.

## Root Cause

`execute_block` treats `SYNCING` as a retryable state but performs the retry in an unbounded loop inside the finalizer actor path. `SYNCING` is not converted into a bounded failure, a cancellable background task, or a mailbox-friendly retry state, so a temporary or persistent local EL condition can monopolize finalizer progress.

## Code

- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1331 enters the finalizer-side execution-layer payload check loop.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1335 logs `execution client returned SYNCING` and retries instead of returning control to the finalizer mailbox.
[...]

PR #202, opened 2026-05-18
fix: buffer blocks in finalizer when EL is syncing without blocking main loop

matthias-wright:
Building on https://github.com/SeismicSystems/summit/pull/167 and https://github.com/SeismicSystems/summit/pull/192.

This addresses https://github.com/SeismicSystems/summit/issues/197.

Changes:
- Removes retries of block execution in unbounded loop when the EL is syncing.
- Introduces buffers for the notarized and finalized blocks. If the EL returns `SYNCING`, the block is buffered and the main loop of the finalizer actor can continue. The block buffers are drained periodically. The interval is `FINALIZER_DRAIN_INTERVAL`, currently set to 5 secs.
- The buffers are unbounded. After either one of them reached `FINALIZER_BUFFERED_BLOCKS_WARN_THRESHOLD`, warnings will be fired through logs and metrics.
- Regression tests are added to verify the correct behavior.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -282,40 +282,73 @@ impl<
             }))
             .await;
 
-        // Send initial forkchoice to the execution client so it knows the chain
-        // head and can start P2P sync. Then wait for sync to complete before
-        // replaying any blocks. Without this, catch-up blocks fail because the
-        // execution client doesn't have them yet.
+        // Send initial forkchoice to the execution client so it knows the
+        // chain head and can start P2P sync.
+        // We do not block the actor here waiting for SYNCING to resolve.
+        //
+        // If the EL is still SYNCING, we fall through and enter the main
+        // mailbox loop immediately. The first finalized or notarized block
+        // we receive goes through `execute_block`, whose own SYNCING-aware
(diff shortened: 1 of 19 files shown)


evonide:
The deadlock fix looks right: when Reth returns SYNCING, the finalizer no longer waits inside the actor loop and can keep answering other messages.

There is one open question regarding the new retry queue though. For finalized blocks this seems naturally limited because syncer waits for acknowledgements before sending too many. However, notarized blocks do not seem to have the same backpressure, and 100 looks like a warning threshold rather than a hard cap.

So does this change the failure mode from “finalizer gets stuck” to “finalizer stays alive, but may keep accumulating notarized blocks while Reth is syncing”? If so, should pending notarized blocks be deduped/bounded, or is there another reason this queue cannot grow into a resource issue during prolonged EL syncing?

matthias-wright:
Agreed, if Reth returns SYNCING for whatever reason, the pending notarized queue would grow indefinitely.
I added dedup for the pending notarized queue, and introduced a hard cap `finalizer_pending_notarized_max` on it. When the cap is reached, a graceful shutdown will be initiated with error logs. The cap can be set via CLI flag and defaults to 1000.

evonide:
LGTM!

Issue #195, opened 2026-05-15
[Security] Critical (attackable): Unauthenticated Genesis RPC Can Replace Startup Genesis


evonide:
## Context

Summit can provision genesis at startup through an RPC path when no local genesis file is present. Genesis defines the chain identity, validator set, execution genesis hash, peer configuration, and initial protocol parameters, so first-boot provisioning is a high-trust boundary.

The intended flow is that an operator supplies the intended genesis before consensus starts, then the node treats that genesis as authoritative for the chain it joins.

## Claim

A remote network caller that can reach the first-boot genesis RPC can call unauthenticated `sendGenesis` before genesis is loaded. Summit writes the caller-controlled genesis content to the configured path and then loads it as the node's authoritative chain identity and validator configuration.

## Impact

An attacker can install a malicious genesis on affected first-boot nodes. That controls chain identity, namespace, execution genesis hash, validator committee, peer addresses, and initial protocol parameters before consensus starts. The result is full bootstrap compromise for affected nodes: they can join an attacker-defined chain or crash on malformed replacement genesis.

## Root Cause

Genesis provisioning exposes an unauthenticated write RPC on a public listener, and startup treats the resulting file as authoritative chain identity without authenticating the writer or binding the file to an expected genesis.

## Code

- https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L169 detects missing genesis and creates the provisioning signal.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L244 starts the genesis RPC before loading genesis.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L260 waits for the genesis RPC signal.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L265 loads whatever file is present at `flags.genesis_path`.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/api.rs#L112 exposes `sendGenesis` without an authentication parameter.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/genesis.rs#L60 writes caller-controlled genesis content to disk.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/genesis.rs#L69 notifies startup after the write.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/builder.rs#L38 binds RPC to `0.0.0.0`.
[...]

PR #203, opened 2026-05-18
fix: bind genesis RPC server to loopback

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/193.

This addresses https://github.com/SeismicSystems/summit/issues/195.

Changes:
- Binds the genesis RPC server to 127.0.0.1.
- Adds a regression test.

I didn't add an authentication token for this endpoint, because this endpoint won't be used until we switch to TEEs. How this endpoint will be authenticated will be decided then.


sebastian-osec:
Looks good to me.

Issue #199, opened 2026-05-18
[Security] Medium (correctness): Recovered Consensus State Loses Frozen Proof Root


sebastian-osec:
## Context

`ConsensusState` maintains mutable consensus fields together with an SSZ Merkle tree and a frozen proof snapshot used for block auxiliary data, including the parent beacon block root. After block execution, Summit captures a state root/proof snapshot so later epoch-transition mutations do not change the root that validators use for proposal construction and verification.

Restart recovery must restore that same frozen snapshot for a node to stay aligned with uninterrupted peers.

## Claim

During a normal restart after an epoch-boundary persisted-state write, a validator can load serialized consensus state whose frozen proof snapshot was omitted. Recovery then rebuilds `state_root` and `proof_tree` from the post-transition live fields, causing restarted validators to derive a different expected `parent_beacon_block_root` than validators that never restarted.

## Impact

Honest validators can disagree on block validity after restart. Proposals built by restarted validators can be rejected by uninterrupted validators because the `parent_beacon_block_root` does not match their expected aux-data root; restarted validators can likewise reject proposals from uninterrupted peers.

This can halt chain progress/finality when enough validator weight restarts in the affected window, because the validator set is split across incompatible parent-beacon-root expectations rather than voting on the same valid proposal stream.

## Root Cause

The frozen proof snapshot fields are runtime-only and skipped during serialization. On decode, Summit default-initializes those fields and calls `rebuild_ssz_tree()`, which assigns `state_root` and `proof_tree` from the current live consensus fields. At an epoch boundary, the finalizer can persist consensus state after live fields have changed from the state that originally produced the frozen proof root, so restart recovery loses the historical root that uninterrupted validators still use.

## Code

- `ConsensusState` keeps the live SSZ tree as runtime state skipped by serialization: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L51
- The frozen proof snapshot fields, including `proof_tree`, `proof_validator_keys`, and `proof_el_block_number`, are also skipped by serialization: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L63
- `capture_state_root()` freezes the current SSZ root/proof state after block execution: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L472
[...]

PR #205, opened 2026-05-19
fix: persist captured proof snapshot across restart

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/170.

This addresses https://github.com/SeismicSystems/summit/issues/199.

Changes:
- Adds a field `captured_bytes` to `ConsensusState` that contains a snapshot of the consensus state (with `captured_bytes` set to `None`). This snapshot is taken in `capture_state_root()`. This snapshot is used to derive the fields `proof_tree`, `state_root`, and `proof_validator_keys`. This ensures that the frozen state root is not lost by a validator restart, especially at the epoch boundary.
- Adds a unit test to verify the behavior.

### DIFF
--- a/application/Cargo.toml
+++ b/application/Cargo.toml
@@ -36,5 +36,5 @@ alloy-rpc-types-engine.workspace = true
 
 [features]
 prom = ["metrics"]
-bench = []
+bench = ["summit-types/bench"]
 permissioned = []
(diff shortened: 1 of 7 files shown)


sebastian-osec:
Looks good to me.

One deployment note: this changes the serialized ConsensusState format by adding proof_el_block_number/captured_bytes trailers. If nodes restart from state encoded before this PR, decode requires a migration/backward-compatible read path

Issue #200, opened 2026-05-18
[Security] Critical (attackable): Fee Recipient Aux Data Is Not Verified


sebastian-osec:
## Context

Summit's application layer constructs and verifies execution payloads while the finalizer supplies auxiliary data such as the expected fee recipient. Honest proposal construction uses this aux value to follow treasury or proposer withdrawal-credential policy when asking the execution engine to build a payload.

Remote block verification is the trust boundary for blocks received from other proposers. It must independently enforce Summit-specific payload policy because the execution engine validates EL rules, but does not know whether Summit expected fees to go to the treasury or to a particular proposer address.

## Claim

A Byzantine proposer running modified proposal code or a modified execution engine can build an otherwise valid execution payload whose actual fee recipient differs from `BlockAuxData.suggested_fee_recipient`. Summit's remote block verification does not compare the payload fee recipient against the expected aux-data recipient, so the policy-violating block can still pass consensus verification.

## Impact

A Byzantine proposer can redirect execution fees away from Summit's treasury or proposer withdrawal-credential policy while still producing a block that honest validators accept. The attacker-chosen recipient becomes the canonical execution payload recipient for that block on all honest nodes that accept and finalize it.

This is a direct value-redirection issue: Summit computes an expected recipient, honest construction uses it, but verification does not enforce it.

## Root Cause

The expected fee recipient is computed and passed into honest payload construction, but it is not checked during remote block verification and is not passed to execution-layer validation as a Summit-specific invariant. `newPayloadV4` validates execution payload correctness, but not Summit's fee-recipient policy.

## Code

- Finalizer computes `suggested_fee_recipient` from treasury address or proposer withdrawal credentials: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L906
- If treasury is zero, the expected recipient falls back to the proposer's withdrawal credentials: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L910
- `BlockAuxData` carries `suggested_fee_recipient`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/lib.rs#L54
- Finalizer includes the expected recipient in aux data: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L974
[...]

PR #206, opened 2026-05-19
fix: validate fee recipient for proposed block

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/191 (which builds on https://github.com/SeismicSystems/summit/pull/190).

This addresses https://github.com/SeismicSystems/summit/issues/200.

Changes:
- Adds a check for the fee recipient to `handle_verify`. This check is only performed, if the `treasury_address` is non-zero. If the `treasury_address` is set to the zero address, then block proposers may chose any fee recipient.
- Adds unit tests for verifying this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6026,6 +6026,8 @@ dependencies = [
 name = "summit-application"
 version = "0.0.3-alpha"
 dependencies = [
+ "alloy-primitives",
+ "alloy-rpc-types-engine",
  "anyhow",
  "commonware-consensus",
  "commonware-cryptography",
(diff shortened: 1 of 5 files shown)


evonide:
This appears to verify the fee recipient when treasury mode is nonzero but leaves the zero-treasury/proposer-recipient case unchecked. Should verification require the payload fee recipient to match the expected proposer withdrawal address in that mode, too?

matthias-wright:
When the treasury address is set to zero, the block proposer may use its own address for the fee recipient (or any other address), so I think this check should cover it.

evonide:
LGTM! This passes the treasury address through aux data and rejects received blocks whose payload pays a different recipient.

Issue #201, opened 2026-05-18
[Security] Critical (attackable): Unpermissioned malformed deposit can drop valid same-entry deposits


evonide:
## Context

Summit consumes execution-layer EIP-7685 request entries to turn deposit data into consensus-side validator credits, admissions, or refunds. A type-`0x00` deposit request entry can contain multiple Summit deposit chunks, so parsing is the boundary between public deposit-contract events and per-deposit handling.

The public deposit contract accepts opaque node and consensus key bytes after checking field lengths, withdrawal-credential format, value bounds, and the supplied deposit-data root. Summit later applies stricter parser rules by decoding the node key as Ed25519 and the consensus key as BLS before any per-deposit admission or refund logic runs.

## Claim

An unpermissioned execution-layer account that can pay the 1 ETH minimum deposit can submit a public deposit-contract call with length-valid, root-consistent key bytes that the contract accepts and emits, but Summit later cannot decode as an Ed25519 node key or BLS consensus key. Reth/EIP-6110 deposit extraction concatenates same-block deposit logs into a single type-`0x00` EIP-7685 deposit request entry, so if the attacker-funded parser-invalid deposit is grouped with victims' valid deposits, Summit discards the whole entry before credit or refund handling.

## Impact

Valid deposits in the skipped entry may already be locked or burned on the execution side, while Summit neither credits validators nor creates refunds. Summit source shows no later replay or refund path for valid chunks inside a parse-failed grouped entry, so those deposits can become uncredited and unrefunded, with direct funds loss or permanent stranding under the documented deposit-contract burn/lock model.

## Root Cause

Deposit entry parsing is all-or-nothing at the grouped entry level. `DepositRequest::try_from_eth_entry_bytes(...).collect()` returns one `Result<Vec<_>, _>`, and `parse_execution_requests` skips the entire execution-request entry on `Err` instead of processing valid chunks independently or rejecting the block.

## Code

- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1477 parses each execution-request entry with `ExecutionRequest::try_from_eth_entry` and only processes requests from the `Ok` branch.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1712 logs and skips the entire entry when parsing returns `Err`.
[...]

PR #207, opened 2026-05-19
fix: handle malformed deposit requests

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/202 (which builds on https://github.com/SeismicSystems/summit/pull/167 and https://github.com/SeismicSystems/summit/pull/192).

This addresses https://github.com/SeismicSystems/summit/issues/201.

Changes:
- Introduces an enum `ParsedExecutionRequest` with variants `Valid(ExecutionRequest)` and `MalformedDeposit(MalformedDepositRequest)`. `ExecutionRequest` is the already existing execution request type, and `MalformedDepositRequest` is a new type for a deposit request with invalid keys.
```rust
pub enum ParsedExecutionRequest {
    Valid(ExecutionRequest),
    MalformedDeposit(MalformedDepositRequest),
}
```
- Arriving execution requests are parsed into `ParsedExecutionRequest` first, and then processed. Malformed deposit requests will be rejected and a withdrawal will be initiated similarly to deposit requests that contain invalid signatures.
- Unit tests are added to ensure that a malformed deposit request cannot drop valid same-entry deposit requests.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -32,7 +32,9 @@ use summit_syncer::Update;
 use summit_types::account::{ValidatorAccount, ValidatorStatus};
 use summit_types::checkpoint::Checkpoint;
 use summit_types::consensus_state_query::{ConsensusStateRequest, ConsensusStateResponse};
-use summit_types::execution_request::{DepositRequest, ExecutionRequest, WithdrawalRequest};
+use summit_types::execution_request::{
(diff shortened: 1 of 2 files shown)


evonide:
LGTM. This PR parses grouped deposit entries per 288-byte deposit chunk so a malformed deposit is isolated/refunded while valid same-entry deposits still reach normal processing.

Issue #204, opened 2026-05-19
[Security] Medium (correctness): Stake-Bound Enforcement Can Nullify Last-Block Exits


sebastian-osec:
## Context

Summit buffers some execution requests that arrive on the last block of an epoch and replays them during later block processing. Active-validator exits use the same withdrawal admission machinery as ordinary withdrawals, including the `has_pending_withdrawal` guard that prevents duplicate pending withdrawals for one account.

Stake-bound enforcement also runs around epoch-boundary protocol-parameter changes and can schedule protocol-induced withdrawals when a validator exceeds the new maximum stake. A full exit that was already accepted at the boundary should not be downgraded or skipped by later stake-bound cleanup for the same validator.

## Claim

During normal epoch-boundary processing, a valid active-validator full exit on the last block of an epoch can be buffered into `pending_execution_requests` while a queued `MaximumStake` reduction takes effect at the same boundary. Stake-bound enforcement can then schedule an excess-stake withdrawal first and set `has_pending_withdrawal`; when the deferred full exit is replayed, withdrawal admission treats it as a duplicate pending withdrawal and skips it.

## Impact

A validator's valid full exit can be converted into only a protocol-induced partial withdrawal. The validator remains active at the new capped balance instead of exiting, so validator-set and withdrawal semantics depend on whether a maximum-stake update shared the same epoch boundary.

This can leave validators active after they submitted a valid exit and can weaken stake-backed committee assumptions for affected validators, but the demonstrated impact is narrower than funds loss, finalized-state divergence, or network-wide liveness failure.

## Root Cause

Last-block active-validator exits are stored as ordinary pending execution requests and later re-enter the normal withdrawal admission path. Epoch-boundary stake-bound enforcement runs before that deferred exit is replayed, and the shared `has_pending_withdrawal` guard does not distinguish a full exit already buffered at the previous boundary from a newly scheduled excess-stake withdrawal.

## Code

- Last-block active-validator withdrawals are buffered into `pending_execution_requests` instead of being applied immediately: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1653
- The deferred withdrawal request is serialized and pushed into pending execution requests, then processing continues without marking the validator for exit: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1664
[...]

PR #208, opened 2026-05-20
fix: prioritize last-block validator exits over stake-bound partial withdrawals

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/188 (which builds on https://github.com/SeismicSystems/summit/pull/187).

This addresses https://github.com/SeismicSystems/summit/issues/204.

Changes:
- On the last block of an epoch, in case the min or max stake was changed, the `pending_execution_requests` will be peeked and validators that have an exit scheduled for the next epoch will be skipped when creating partial withdrawals. This ensures that a partial withdrawal won't block the validator exit.
- Adds a unit test to verify the behavior.


sebastian-osec:
Looks good to me.

Issue #209, opened 2026-05-22
[Security] High (attackable): PREVRANDAO is unconstrained by Summit verification


evonide:
## Context

Summit proposers build execution payloads through the Engine API, then validators verify the Summit block before voting on it. The execution payload is committed into the Summit block digest, so all validators agree on the same payload bytes.

One of those payload fields is `prev_randao`, which is exposed to EVM contracts through the `PREVRANDAO` opcode. That value should come from a consensus-defined randomness source or another explicitly defined Summit policy, not arbitrary proposer choice.

## Claim

A malicious scheduled proposer running modified Summit or EL-facing code can include an otherwise valid execution payload with a chosen `prev_randao` value. Summit verification accepts the payload bytes without comparing `prev_randao` to any consensus-derived expected value, so the attacker-controlled value can pass into the canonical payload.

## Impact

Contracts that use `PREVRANDAO` for randomness can receive a proposer-chosen value instead of unbiased consensus randomness. This can make lotteries, mints, games, ordering decisions, or other entropy-sensitive EVM logic predictable or biasable.

The payload hash only proves validators agreed on the chosen bytes. It does not prove the embedded randomness value was valid under Summit’s consensus rules.

## Root Cause

Summit sets `prev_randao` to zero during honest payload construction, but remote block verification does not enforce zero or any other expected consensus-derived value. The verifier checks other block/header/auxiliary fields, but never checks `payload.prev_randao`.

## Code

- https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L101 sets honest payload construction `prev_randao` to zero.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L551 commits the execution payload returned by the proposer into the Summit block.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L51 hashes the full execution payload bytes into `payload_hash`.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L598 verifies parent, height, timestamp, state root, checkpoint, validator deltas, and withdrawals, but does not compare `payload.prev_randao` to an expected value.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L149 replays the payload through `new_payload_v4`, but does not pass any Summit-derived expected randomness value.

## Related Issues/PRs

[...]

PR #263, opened 2026-05-22
fix: verify PREVRANDAO

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/206 (which builds on https://github.com/SeismicSystems/summit/pull/191 and https://github.com/SeismicSystems/summit/pull/190).

This addresses https://github.com/SeismicSystems/summit/issues/209.

Changes:
- In `handle_verify`, checks that the `prev_randao` is zero.
- Adds unit tests.


sebastian-osec:
This fix looks good to me.

Issue #210, opened 2026-05-22
[Security] High (attackable): Genesis max deposits cap can stall finalization


evonide:
## Context

Summit genesis supplies the initial protocol parameters used by finalizer logic before any runtime protocol-parameter update path can constrain them. One of those parameters, `max_deposits_per_epoch`, controls how much deposit work the finalizer attempts at epoch boundaries.

Runtime protocol-parameter updates cap this value, but genesis is loaded first and copied into consensus state as the chain's starting configuration.

## Claim

A remote first-boot genesis RPC caller can supply genesis bytes with `max_deposits_per_epoch` above the runtime protocol maximum. Summit accepts the unchecked genesis value, stores it in consensus state, and later penultimate-block processing loops over the full configured count even when the deposit queue is empty.

## Flow

1. A node starts without a local genesis file and waits for first-boot genesis provisioning.
2. A caller that reaches the first-boot genesis RPC sends attacker-controlled genesis bytes.
3. Startup loads that genesis and copies `max_deposits_per_epoch` into consensus state without applying the runtime cap.
4. At penultimate-block processing, the finalizer loops over the full configured deposit cap.
5. If the deposit queue is empty, `pop_deposit()` returns `None`, but the loop continues instead of breaking.
6. With an effectively unbounded value, affected validators can spend the consensus-critical penultimate block in a non-terminating loop.

## Impact

This gives the first-boot genesis caller a concrete path from public/bootstrap genesis provisioning into deterministic finalization liveness failure for nodes that accepted that genesis.

The same unchecked value can also be introduced by a trusted genesis author or faulty provisioning flow, but the critical path is the public first-boot handoff: attacker-controlled genesis is consumed as authoritative startup state and later drives finalizer behavior.

## Root Cause

Genesis validation does not enforce the same `max_deposits_per_epoch` upper bound that runtime protocol-parameter updates enforce. Separately, penultimate-block deposit processing does not break when `pop_deposit()` returns `None`, so an empty queue still causes the loop to continue for the full configured cap.

## Code

- https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L55 exposes `max_deposits_per_epoch` as a raw `u64` genesis field.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L134 validates unrelated genesis fields but does not enforce the runtime maximum for `max_deposits_per_epoch`.
[...]

PR #264, opened 2026-05-22
fix: enforce protocol-param bounds at genesis and break empty deposit loop

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/210.

Changes:
- Adds checks for max_deposits_per_epoch, max_withdrawals_per_epoch, and observers_per_validator values from the genesis file.
- Adds a break the deposit loop when pop_deposit() returns None.


sebastian-osec:
This fix looks good to me.

Issue #260, opened 2026-05-22
[Security][tentative] Medium (correctness): State-Proof Batches Drop Missing Keys


evonide:
## Context

Summit's state-proof RPC lets external consumers request Merkle proofs for consensus-state fields and verify those proofs against a returned state root. Batched requests are intended to preserve the caller's association between each requested key and each returned proof.

This finding concerns the batch response shape when some requested keys cannot be proven. The normal flow is that every requested key should either produce a proof or an explicit per-key failure so consumers do not have to infer which proof corresponds to which request.

## Claim

`getStateProof` accepts batched key requests, but proof generation uses `filter_map`, so any key whose proof returns `None` is silently omitted. The response then contains a shorter proof list with no echoed key or per-key status, allowing positional consumers to associate later proofs with the wrong requested fields.

A batched `getStateProof` request mixes valid keys with absent or out-of-range keys. Summit silently drops the missing entries instead of preserving one result per request, so a downstream consumer that relies on positional association can map returned proofs to the wrong requested fields.

## Flow

The path is reachable through the RPC state-proof endpoint whenever a batch mixes provable keys with keys whose proof generator returns `None`. The start condition is ambiguous client/API semantics, not an RPC caller directly gaining authority by attacking its own request; impact requires a positional downstream consumer because the response does not echo request keys or statuses.

## Impact

A consumer that maps returned proofs to requested keys by position can be shifted by an attacker-controlled missing key. Each Merkle proof may verify against the returned root while being interpreted as proof for a different requested field, misleading external proof consumers, accounting systems, or contracts that rely on batched responses.

## Root Cause

Batch proof generation uses `filter_map`, which removes failed proof requests instead of preserving request cardinality with an explicit missing result or error.

## Code

- Docs describe `getStateProof` as returning an `SszProof` for each key: https://github.com/SeismicSystems/summit/blob/ed2c5c8/docs/ssz-merklization.md#L372.
- RPC parses all requested keys and forwards the full vector: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L410.
- Finalizer generates proofs with `filter_map`, dropping keys whose proof function returns `None`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1107.
[...]

PR #267, opened 2026-05-27
fix: update state proof batch result alignment

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/260.

  Changes:

  - Preserve one internal proof result per requested key with Option<SszProof>.
  - Return keyed RPC results[] entries with proof or error.
  - Add a regression test for mixed valid/missing batched keys.
  - Update docs and the e2e proof verifier for the new response shape.
  - Adds unit test `test_generate_state_proof_preserves_batch_cardinality_for_missing_keys` .

### DIFF
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -371,19 +371,32 @@ The state root appears on-chain in EL block `el_block_number + 1`.
 
 ### `getStateProof`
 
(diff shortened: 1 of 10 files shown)


sebastian-osec:
This looks good to me. The PR preserves one keyed result per requested proof key, so missing keys are reported explicitly instead of shifting the batch response.

Issue #252, opened 2026-05-22
[Security][tentative] Medium (correctness): P2P Message Size Can Prevent Large Block Sync


evonide:
## Context

Summit peers use Commonware P2P channels for block propagation and resolver backfill. Lagging or checkpoint-joining nodes rely on single-message block and certificate responses to repair finalized data gaps, while `max_message_size_bytes` limits the largest payload those channels will send.

## Claim

Summit sends full blocks and full `(certificate, block)` backfill responses as single Commonware P2P messages. It does not validate `max_message_size_bytes` against the largest valid block plus resolver and certificate overhead, so valid finalized data can be larger than the only backfill message format used to serve it.

A `max_message_size_bytes` configuration below the size of valid finalized block or `(certificate, block)` data can leave Summit's single-message backfill response larger than the transport cap. Lagging, restarting, or checkpoint-joining peers then request that finalized data through normal resolver backfill, but every single-message response exceeds the cap and the peers remain unable to repair the gap. A malicious proposer path would require separately proving that valid execution payloads in the target deployment can exceed the configured cap.

## Flow

The path is reachable whenever finalized block repair or boundary-finalization backfill serves full responses as single P2P messages. It requires a valid block or `(certificate, block)` response exceeding the configured `max_message_size_bytes`, so adversarial proposer reachability depends on valid execution payloads being able to exceed the configured cap in the target deployment.

## Impact

A lagging, restarting, or checkpoint-joining node can learn that a height is finalized but be unable to obtain the corresponding block or boundary finalization response if the encoded response is larger than the P2P message cap. The resolver retries, but every serve response fails the same size check, so archive gap repair and finalized-block dispatch can stall until manual intervention or an out-of-band block import.

## Root Cause

Summit exposes a P2P message-size limit without validating it against the largest valid block/backfill response and without chunking resolver responses.

## Code

- Genesis exposes only a message-size knob, not a block-size relationship, and validates unrelated fields: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L35, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L135.
[...]

PR #268, opened 2026-05-27
fix: enforce max size of max_message_size_bytes/2 for proposed blocks

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/252.

Changes:
- Reject blocks whose size exceeds `max_message_size_bytes/2` in `handle_verify` and `handle_certify`.
- Set `max_message_size_bytes` to 10MB. This ensures that `(certificate, block)` won't exceed `max_message_size_bytes`.


sebastian-osec:
This PR does add block-size validation in `application/src/actor.rs`: proposed/verified/certified blocks are rejected when `block.encode_size()` is at least `max_message_size_bytes / 2`. That is definitely a real guard and reduces the risk.

My remaining concern is that #252 is specifically about the actual single-message resolver/backfill payloads. On this branch, `syncer/src/actor.rs` still serves `block.encode()`, `(finalization, block).encode()`, and `(notarization, block).encode()` as single responses, and I do not see those actual encoded response sizes checked against `max_message_size_bytes`.

So the fix seems to rely on the invariant that certificate/resolver/P2P overhead always fits in the remaining half of the message budget. Could we either validate the actual encoded resolver responses directly, or document/prove that this overhead is always bounded below `max_message_size_bytes / 2` for the supported validator/config limits? A regression test for a lagging/checkpoint-joining node requesting a near-limit finalized block would also help confirm the original backfill failure mode is fixed.

matthias-wright:
Update(cbd7cef):
- Added comment with some Math that justifies why the overhead will always be below `max_message_size_bytes / 2` for any realistic validator counts

sebastian-osec:
Thanks, the added explanation makes sense to me.

Given that proposed/verified/certified blocks are now capped at `max_message_size_bytes / 2`, `Request::Block` responses should always fit, and the remaining half should cover the finalized/notarized certificate overhead for realistic validator counts. The overhead being one aggregate signature plus a signer bitmap is the key point, since it scales as roughly `ceil(N / 8)` rather than `N * signature_size`.

So I think the substantive issue is addressed.

Optional hardening nit: this is still documented as a size proof rather than checked on the actual encoded resolver responses, so if Summit ever supports extremely large validator sets or changes certificate encoding/signature aggregation, this assumption should be revisited. A small regression/unit test around the near-limit `(certificate, block).encode_size()` case would also make the invariant more explicit.

matthias-wright:
Update(366f72a):
- Test: assert (certificate, block) backfill response fits the P2P cap

Issue #225, opened 2026-05-22
[Security][tentative] High (attackable): Public State-Proof RPC Can Stall the Finalizer


evonide:
## Context

Summit's public RPC layer exposes state-query methods that route into the finalizer actor. The finalizer also performs consensus-critical work such as processing finalized and notarized blocks, preparing auxiliary data, and serving proofs over its current consensus state.

The intended boundary is that public read-only RPC work should not monopolize the actor loop that advances local finalization. `getStateProof` lets callers choose proof keys, and some proof paths depend on consensus-state collection lookups.

## Claim

`getStateProof` accepts caller-controlled key batches and performs proof generation synchronously on the finalizer actor's main event loop. Validator proof generation scans the validator key list per requested key, allowing large batches to monopolize the finalizer.

An unauthenticated remote RPC caller can send large `getStateProof` key batches to an exposed node; the request is forwarded to the finalizer actor, which generates proofs synchronously on its main loop, delaying finalized/notarized block handling and other consensus-critical finalizer work.

## Flow

The path is reachable on nodes exposing the public RPC server. The attacker only needs to send `getStateProof` requests within the configured request-size limit.

## Impact

An unauthenticated remote caller can submit large batches of validator proof keys and force repeated scans and proof allocations inside the finalizer actor. While that work runs, the node is delayed in processing finalized/notarized blocks and aux-data requests, which can degrade or halt local consensus progress.

## Root Cause

Expensive proof generation is executed synchronously inside the finalizer actor's main event loop, and validator proof lookup scans the validator list for each requested key.

## Code

- The RPC server binds to `0.0.0.0` by default and allows 10 MiB request/response bodies: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/builder.rs#L35-L38, https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/lib.rs#L52.
- `getStateProof` accepts a `Vec<String>` from remote callers: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/api.rs#L100.
- The RPC handler parses the full key vector and forwards it to the finalizer mailbox: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L410.
- The finalizer handles `QueryState` in the same select loop as other finalizer messages: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L381.
[...]

PR #269, opened 2026-05-27
fix: protect finalizer loop from expensive RPC requests

matthias-wright:
Addresses: https://github.com/SeismicSystems/summit/issues/225

  Changes:
  - Add a dedicated finalizer state-query mailbox for RPC state queries.
  - Prioritize the main finalizer mailbox with `select_biased!`, leaving query work last.
  - Move `GenerateStateProof` proof generation off the finalizer actor loop by snapshotting frozen proof data and spawning proof work on a blocking/shared task.
  - Store frozen proof tree and validator keys behind `Arc` so proof requests clone cheap handles.
  - Add `getStateProof` request limits: max 128 keys, max computed cost 512, with RPC errors 3005/3006.
  - Update RPC wiring and tests to use the query mailbox.
  - Add tests for oversized proof key batches and excessive proof cost.
  - Update finalizer tests for the new constructor return value.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -1910,7 +1910,7 @@ impl<
                     el_block_number,
                 });
             }
-            ConsensusStateRequest::GenerateStateProof(keys) => {
+            ConsensusStateRequest::GenerateStateProof(keys, permit) => {
                 let proof_tree = self.canonical_state.proof_tree_snapshot();
                 let validator_keys = self.canonical_state.proof_validator_keys_snapshot();
                 let root = self.canonical_state.get_state_root();
@@ -1929,6 +1929,18 @@ impl<
                             el_block_number,
                             proofs,
                         });
+                        // Release the rpc concurrency permit (if any) only now
+                        // that the proof work has actually finished. The permit
+                        // must be consumed inside this detached task because
+                        // moving it here is what ties the in-flight-proof count
+                        // to real work. If it instead dropped on the rpc handler
+                        // future, a caller could connect, wait for the spawn,
+                        // disconnect, and repeat to pile up proof tasks under a
+                        // slot count that reads as idle. The explicit drop also
+                        // forces the move closure to capture it rather than
+                        // dropping it on the finalizer loop when the match arm
+                        // ends.
+                        drop(permit);
                     });
             }
         }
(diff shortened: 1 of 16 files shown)


sebastian-osec:
This looks like it fixes the core finalizer-mailbox isolation problem: public state queries now use a separate query channel, the main finalizer mailbox is prioritized, and proof generation snapshots the frozen proof tree instead of doing the expensive proof work inline on the finalizer actor loop.

Two concerns I have:

1. Accepted `GenerateStateProof` requests still spawn independent shared `state-proof` tasks, and I do not see a global/semaphore cap on concurrent proof generation. The per-request key/cost limits bound each request, but a remote caller can still create many accepted proof tasks. Can this be bounded so the pressure is not just moved from the finalizer loop to the shared task pool?

2. This branch still builds proofs with `filter_map(...).collect::<Vec<SszProof>>()`. When this composes with #267, please preserve the keyed `Vec<Option<SszProof>>` / one-result-per-requested-key behavior so missing keys cannot shift result alignment again.

matthias-wright:
Update(cadb8f9):
- Added cap on concurrent proof generation

sebastian-osec:
One remaining question: is the permit lifetime tied to the actual spawned proof-generation task, or only to the RPC handler future?

Right now the slot appears to be held in `get_state_proof()` while awaiting `state_query.generate_state_proof(...)`, but the finalizer then spawns detached `state_proof` work. If a client disconnects or the RPC future is cancelled after the finalizer has accepted the request and spawned the proof task, the `StateProofSlot` may be dropped and capacity freed while the proof task continues running and eventually fails to send on the oneshot.

If that cancellation path is possible in jsonrpsee/commonware, then the cap bounds in-flight RPC waits but not necessarily actual proof-generation work. Could we either move the permit into the spawned proof task, or otherwise ensure the slot is released only when proof generation finishes?

matthias-wright:
Update(dc93fb2):
- Tie state-proof concurrency permit to the proof task, not the RPC future

sebastian-osec:
Thanks, I think the latest commit addresses my remaining permit-lifetime concern.

The state-proof slot is now acquired at RPC admission, moved through `GenerateStateProof(..., Some(permit))`, and dropped inside the detached `state_proof` task after proof generation finishes.

I re-checked the broader #225 path as well. Public RPC now uses the dedicated `ConsensusStateQuery` mailbox, the finalizer prioritizes the main mailbox with `select_biased!`, proof generation runs off-loop against a frozen snapshot, and the key-count / cost / concurrency / cancellation / positional-result tests cover the important cases.

Small nit: the doc comment above `ConsensusStateQuery::generate_state_proof` appears duplicated.

Optional follow-up: validator proof lookup still scans the validator-key list, so indexing validator positions could be a useful performance improvement later if large validator sets make this noticeable.

Issue #246, opened 2026-05-22
[Security][tentative] Medium (correctness): Finalizer Persists Validator Transition Fields Before Cleanup


evonide:
## Context

The finalizer maintains the canonical consensus state that is committed to disk and reloaded after restart. Around an epoch transition, validator-set deltas are needed transiently for headers and aux data, but the durable post-transition state should match what continuously running peers keep in memory.

This finding concerns the boundary between durable recovery state and transition-only fields. If those fields are committed before cleanup, restarted nodes can recover a different state shape than peers that crossed the same epoch without restarting.

## Claim

The finalizer commits transition-only `added_validators` and `removed_validators` before cleanup, so restart recovery can resurrect state that live peers already discarded.

An operator restart or local fault after an epoch-boundary DB commit can make a node reload transition-only `added_validators` or `removed_validators` fields that live peers already cleared, leaving the restarted node to propose or verify from stale persisted state.

## Flow

The path requires an epoch transition with `added_validators` or `removed_validators`, a successful state commit, and a local restart or recovery from that committed DB state after live peers would have cleared the transition fields. Remote messages, genesis file values, and post-launch protocol inputs are not the claimed way to force an honest node to restart at this precise point.

## Impact

A node restarted after a clean DB commit may propose or verify with stale validator-transition fields. Around validator-set-changing epochs, this can produce header or state-root divergence from non-restarted nodes and cause local consensus stalls.

## Root Cause

The finalizer persists durable `ConsensusState` before clearing fields that are intended to be transition-only header or aux-data inputs.

## Code

- Finalizer stores and commits consensus state at the epoch boundary: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L622, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L637.
- The cleanup happens only after that commit: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L647.
- `added_validators` and `removed_validators` are serialized into DB-backed `ConsensusState`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L1028, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L1039.
[...]

PR #270, opened 2026-05-27
fix: persist cleaned epoch transition state before ack

matthias-wright:
Addresses: https://github.com/SeismicSystems/summit/issues/246 and https://github.com/SeismicSystems/summit/issues/325

Changes
 - Move finalized-block ack after epoch-boundary state persistence.
 - Clear `added_validators` / `removed_validators` before storing next-epoch consensus state.
 - Add a restart regression test that verifies persisted transition deltas are cleared and the active validator set is correct after reload.
 
 Note: once this and https://github.com/SeismicSystems/summit/pull/382 are on the same base, we should add a finalizer test that injects a commit failure at an epoch boundary and asserts the syncer ack is withheld (the Exact waiter resolves Err/canceled) and the node shuts down without durably advancing the epoch.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -916,15 +916,9 @@ impl<
         }
 
         let new_height = block.height();
+        let current_epoch = self.canonical_state.get_epoch();
         self.height_notify_finalized_up_to(new_height, block_digest);
-        ack_tx.acknowledge();
-        info!(
-            new_height,
-            epoch = self.canonical_state.get_epoch(),
-            "executed block"
-        );
 
-        let new_height = block.height();
         let mut epoch_change = false; // Store finalizes checkpoint to database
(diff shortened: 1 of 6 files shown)


sebastian-osec:
This fixes the main #246 persistence issue: transition deltas are cleared before the next-epoch consensus state is stored and committed. One follow-up concern: the finalized-block ack now happens after the oracle/orchestrator epoch-transition reports as well as after DB commit. Is that intended? If oracle/orchestrator reporting can block, this moves external backpressure onto the finalized-block ack path even after durable persistence has succeeded.

matthias-wright:
Update(7077fd6):
- Send finalized block ack before the oracle/orchestrator reports.

sebastian-osec:
This fix looks good to me.

Issue #230, opened 2026-05-22
[Security][tentative] High (correctness): Protocol Stake Bounds Can Become Empty


evonide:
## Context

Summit protocol parameters define the stake interval that controls which validator balances are allowed to enter or remain in the active validator set. Genesis loading and later protocol-parameter updates are expected to preserve a usable interval so deposits, top-ups, and forced-withdrawal handling can evolve the committee safely.

## Claim

Summit validates `validator_minimum_stake` and `validator_maximum_stake` independently at genesis and during protocol-parameter updates, without enforcing `validator_minimum_stake <= validator_maximum_stake`. This allows the active stake interval to become empty.

A genesis/config author or an authorized protocol-parameter path can set minimum stake above maximum stake. Ordinary attacker reachability is not established by this record alone.

## Flow

Reachability is config/authority-only: genesis loading accepts an inverted interval, and accepted protocol-parameter changes can independently apply `MinimumStake` and `MaximumStake` values that leave `validator_minimum_stake > validator_maximum_stake`. A public-attacker path is not proven; the previously considered type `0xFF` protocol-parameter path is documented as owner-gated.

## Impact

If `minimum_stake > maximum_stake`, every possible validator balance is outside the valid range. New deposits cannot join, top-ups cannot restore validity, and stake-bound enforcement can treat all existing validators as invalid at the next boundary. This can halt validator-set evolution or remove the active committee depending on timing.

## Root Cause

The protocol validates stake bounds independently and never checks the cross-field invariant that the minimum must be less than or equal to the maximum.

## Code

- `MinimumStake` parses any `u64` value: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/protocol_params.rs#L42.
- `MaximumStake` parses any `u64` value: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/protocol_params.rs#L54.
- Genesis validation does not check `validator_minimum_stake <= validator_maximum_stake`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L135.
- Protocol changes apply minimum stake independently: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L702.
- Protocol changes apply maximum stake independently: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L707.
[...]

PR #271, opened 2026-05-27
fix: reject inverted validator stake bounds

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/230.

Changes:
  - Reject genesis and decoded consensus states with `validator_minimum_stake > validator_maximum_stake`
  - Validate protocol-parameter batches against the final prospective stake interval before applying them
  - Skip and drain invalid protocol-parameter batches without shutting down the finalizer

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -878,7 +878,13 @@ impl<
             }
 
             // Apply protocol parameter changes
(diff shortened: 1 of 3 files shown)


sebastian-osec:
This looks good to me. The PR rejects inverted stake bounds at genesis, consensus-state decode, and protocol-parameter application time, and it drains invalid queued parameter batches so the finalizer does not keep retrying the same unusable update.

Issue #243, opened 2026-05-22
[Security][tentative] Medium (correctness): Epoch Transition Root Is Captured Before Boundary Mutations


evonide:
## Context

Summit's finalizer executes blocks, persists consensus state, advances epoch-boundary fields, and supplies auxiliary data that the application layer uses when constructing or verifying the next execution payload. The state root exposed through this aux data is meant to commit to the consensus state that proof consumers and downstream verification associate with the parent block.

At an epoch boundary, the finalizer mutates protocol parameters, committee membership, epoch counters, and epoch-genesis bookkeeping after processing the boundary block. The disrupted invariant is that the first post-boundary block should advertise a root that reflects those transition mutations rather than the snapshot from before they ran.

## Claim

`execute_block` captures the state/proof root before applying epoch-boundary mutations, then the next block's aux data reuses that pre-transition root. The first block of the new epoch can therefore advertise a parent root that omits the protocol-parameter, committee, epoch, epoch-genesis, and transition-cleanup updates from the finalized boundary block.

A normal proposer after an epoch boundary can produce the first block of the new epoch using aux data whose parent root was captured before the boundary mutations.

## Flow

The path requires a finalized last block of an epoch where `execute_block` first captures the root, then finalization applies protocol parameter, committee, epoch, or epoch-genesis mutations, followed by proposal/verification of the next block. The start condition is not an arbitrary post-boundary state: it specifically depends on the finalizer's frozen `state_root`/proof tree remaining from the just-executed boundary block while the live consensus state has advanced.

## Impact

The first block of a new epoch can be signed by a committee and state that have advanced while its advertised parent root still represents the pre-transition snapshot. External proof consumers get a one-block stale root for validator-set and epoch state, and the block does not commit to the committee transition that authorized it.

## Root Cause

The finalizer captures the state root before epoch-boundary side effects, then later proposal aux data reuses the state root without proving that it includes the boundary transition.

## Code

- `execute_block` normalizes forkchoice and captures the root: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1451, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1460.
- Epoch-boundary finalization applies protocol parameter changes and committee updates after block execution: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L544.
[...]

PR #272, opened 2026-05-27
fix: recapture state root after epoch transitions

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/243.

Changes:
  - Re-capture the finalizer state/proof root after epoch-boundary transition cleanup
  - Ensure first post-boundary aux data uses the post-transition consensus state root
  - Add a regression test covering aux data after an epoch boundary

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -1041,6 +1041,16 @@ impl<
                 self.canonical_state.clear_removed_validators();
             }
 
+            // Re-capture the state/proof root AFTER the epoch-transition mutations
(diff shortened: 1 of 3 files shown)


sebastian-osec:
Looks good to me. This fixes the in-memory aux-data root issue for #243 by recapturing after epoch-transition cleanup. One merge-order concern: this needs to compose with #270 so the final order is cleanup -> recapture post-transition root -> persist/commit cleaned state. If the combined branch persists before cleanup/recapture, the durable state/root story can still be inconsistent even though this PR fixes the live aux-data root.

Issue #234, opened 2026-05-22
[Security][tentative] High (attackable): WithdrawalQueue Deserialization Accepts Inconsistent Indexes


evonide:
## Context

Summit's withdrawal queue tracks pending withdrawals with redundant indexes: a map of withdrawals, a schedule by epoch, and the next execution-layer withdrawal index. Checkpoint and persisted-state decoding must restore these structures with the same invariants the live queue maintains.

This finding concerns deserialization at the startup or checkpoint-import boundary. The normal flow is that restored queue indexes agree with one another before block validation and commit logic consume them.

## Claim

`WithdrawalQueue` deserialization trusts the withdrawal map, per-epoch schedule, map keys, and `next_index` as independent fields instead of enforcing the live queue invariants between them. A restored inconsistent queue can validate with filtered withdrawal data, panic during commit, or reuse an EL withdrawal index after restart.

A malformed checkpoint supplier, compromised deployment artifact, or persisted consensus-state source can encode inconsistent withdrawal-queue indexes that decode successfully.

## Flow

The path is reachable during consensus-state or checkpoint deserialization. It requires serialized withdrawal-queue data with mismatched map, schedule, or `next_index` values; ordinary P2P blocks, EL withdrawal requests, and live queue operations are not the claimed ingress for creating these inconsistent indexes on an already-running honest node.

## Impact

A malformed but accepted checkpoint or persisted consensus state can install withdrawal-queue indexes that different parts of Summit interpret differently. Validators can agree on a last-block withdrawal list derived from the filtered queue, then fail during commit when the unfiltered schedule refers to a missing map entry.

Other inconsistent forms can make commit logic update or clear the wrong validator account, or reuse an execution-layer withdrawal index after restart. The practical impact is checkpoint/import-time consensus-state corruption that can stop finalization or produce non-canonical withdrawal handling for nodes that accepted the malformed state.

## Root Cause

Deserialization trusts redundant withdrawal-queue indexes instead of rebuilding or cross-validating their invariants.

## Code

- `WithdrawalQueue::write` serializes the withdrawal map and the schedule separately: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/withdrawal.rs#L341.
- `WithdrawalQueue::read_cfg` reads map entries without checking the map key equals `PendingWithdrawal.pubkey`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/withdrawal.rs#L378.
[...]

PR #273, opened 2026-05-28
fix: validate withdrawal queue invariants on decode

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/234.

Changes:
- Validate WithdrawalQueue invariants during decode so checkpoint/persisted-state import rejects inconsistent withdrawal maps, schedules, and stale next_index values.


matthias-wright:
These changes are contained in https://github.com/SeismicSystems/summit/pull/276

Issue #248, opened 2026-05-22
[Security][tentative] Medium (attackable): Last-Block Withdrawal Can Be Dropped By Same-Block Deposit


evonide:
## Context

Summit handles validator deposits and withdrawals as ordered execution requests inside finalizer state transitions. Last-block withdrawals are special because exiting an active validator at an epoch boundary can affect the next epoch's validator-set transition data, so the implementation defers that withdrawal path.

The normal invariant is that a valid withdrawal request either marks the validator as exiting or remains pending until it can be applied. A later same-block deposit for the same validator should not erase an already accepted withdrawal transition.

## Claim

When an active validator withdraws on the last block of an epoch, Summit defers the withdrawal into pending execution requests and continues before marking the account as pending withdrawal. If the same block's execution-request vector then contains a later same-validator deposit or top-up, that deposit can set `has_pending_deposit`. On the next block, the deferred withdrawal is consumed first, sees `has_pending_deposit`, and is skipped without being requeued.

A noncanonical execution-request payload, custom EL behavior, or any path that lets Summit receive request entries ordered as withdrawal type `0x01` before same-validator deposit type `0x00` in the last block of an epoch can cause the deferred withdrawal to be skipped later. The previously assumed "normal user can order withdrawal before deposit through canonical EL payloads" start condition is not established because EIP-7685 orders request items by ascending type for the commitment and Summit expands each typed entry into homogeneous requests.

## Flow

The validator must be active, the withdrawal must appear on the epoch's last block, and Summit must process a later same-block deposit/top-up for the same validator before deferred requests are processed in the next block. This means the vulnerable ordering requires a noncanonical or custom execution-request source that preserves withdrawal-before-deposit entry order; ordinary Reth-produced EIP-7685 payload ordering should not be assumed to provide that ordering from user transaction order alone.

## Impact

A finalized withdrawal request can be permanently dropped if a valid same-block deposit/top-up for the same validator follows it in execution-request order. The validator remains staked instead of exiting, while the later deposit path can proceed. This breaks funds/lifecycle correctness for withdrawal requests at epoch boundaries.

## Root Cause

The last-block withdrawal branch defers the request and `continue`s before marking `has_pending_withdrawal` or moving the balance, while later deposit handling can set `has_pending_deposit`; deferred withdrawal processing then treats that flag as a terminal skip.

## Code

[...]

PR #274, opened 2026-05-28
fix: preserve deferred last-block withdrawals

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/248.

Changes:
  - Adds explicit `ExecutionRequestOrigin` for current vs deferred request handling
  - Marks last-block deferred withdrawals as pending before buffering them
  - Allows deferred withdrawal replay through its own pending-withdrawal guard
  - Adds regression coverage for withdrawal-entry-before-deposit-entry ordering
  - Adds test `test_last_block_withdrawal_before_same_block_deposit_not_dropped` for verification


sebastian-osec:
LGTM. The accepted last-block withdrawal is now marked pending before deferral, and only deferred replay bypasses the pending-withdrawal guard, so a later same-block deposit no longer causes the withdrawal to be skipped or dropped.

Issue #229, opened 2026-05-22
[Security][tentative] High (attackable): Process-Time Deposit Refunds Can Merge Later Withdrawals Into a Stale Address


evonide:
## Context

The finalizer handles validator deposits by parsing execution requests, creating or updating validator accounts, and queuing refunds or withdrawals when a request cannot become active stake. Withdrawal queue entries are keyed by validator node pubkey and carry the destination address that should receive the emitted funds.

Rejected deposits and later authorized withdrawals are meant to remain tied to their own withdrawal credentials. Queue merging is a convenience for combining amounts for the same validator pubkey, but it must not let an old refund destination become the destination for a later valid account's withdrawal.

## Claim

Process-time invalid new-validator deposits enqueue refunds under the real validator pubkey instead of a refund-only key. The reachable process-time invalidation case is a deposit that passed initial verification but later fails revalidation because stake bounds changed before deposit processing. Because `WithdrawalQueue` merges entries by pubkey while preserving the first address and epoch, a later valid withdrawal for the same pubkey can be merged into the stale refund entry and paid to the old refund address.

The narrow trigger is that the same-pubkey refund remains pending after the account is removed, then the node pubkey is reused for a later valid account whose withdrawal is queued before the stale refund is emitted.

## Flow

The stale refund must be created by the process-time revalidation branch, which requires stake bounds to change after deposit parsing and before deposit processing. The refund must then remain pending in the withdrawal queue while the same node pubkey is accepted for a later account, and a later authorized withdrawal for that account must be queued before the stale refund entry is emitted. The depositor alone does not force the process-time invalidation; the missing start condition is the intervening stake-bound change.

## Impact

If a node pubkey is reused after a process-time invalid deposit refund is queued but before it is emitted, a later legitimate withdrawal can merge into the stale refund. The full valid validator balance can be paid to old withdrawal credentials rather than the current account's authorized withdrawal address.

## Root Cause

Process-time invalid new-validator refunds use the real validator pubkey as the queue key, and `WithdrawalQueue::push_request` merges entries only by pubkey while preserving the first entry's address.

## Code

- Parse-time rejected deposits use synthetic refund keys: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1556.
- Process-time new-validator refunds use `validator_pubkey: node_pubkey_bytes`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1756.
[...]

PR #275, opened 2026-05-28
fix: isolate process-time deposit refunds

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/229.

Changes:
- Key process-time new-validator deposit refunds with synthetic refund-only keys
- Adds regression test `test_process_time_invalid_new_validator_refund_does_not_merge_with_reused_pubkey_withdrawal`


sebastian-osec:
This fix looks good to me.

Issue #226, opened 2026-05-22
[Security][tentative] High (attackable): Invalid Deposit Refunds Can Starve the Withdrawal Queue


evonide:
## Context

Summit processes execution-layer deposit requests in the finalizer and may refund deposits that cannot be admitted as validators. Refunds and validator exits are both represented as withdrawal requests, and each epoch emits only a configured number of ready withdrawals.

The normal flow should keep rejected deposit traffic from consuming the same scarce withdrawal capacity needed for legitimate exits. Deposit admission limits and withdrawal emission limits are separate controls, so their interaction determines whether invalid deposits can create exit-queue pressure.

## Claim

Deposits rejected after execution-request entry parsing can create refund withdrawals before deposit queue admission and outside `max_deposits_per_epoch`. Those refunds then share the same capped epoch withdrawal queue as validator exits, so many refundable invalid deposits can consume withdrawal slots mainly for gas and temporary lockup cost.

An execution-layer depositor can submit many parseable but invalid deposit requests with refund credentials, causing refund withdrawals to consume consensus withdrawal slots when they are emitted.

## Flow

The path is reachable only for deposits that pass Summit's execution-request entry parser, including fixed-size deposit chunks and decodable Ed25519/BLS public keys, and still contain parseable withdrawal credentials when they reach the finalizer refund path. Fully malformed entries or malformed deposit public keys are dropped by parsing before refund construction. The attack requires enough invalid deposits to exceed available `max_withdrawals_per_epoch` capacity alongside honest withdrawal demand.

## Impact

A low-cost stream of refundable invalid deposits can monopolize or heavily occupy the terminal-block withdrawal slots across many epochs. Honest validator exits, stake-bound withdrawals, and normal refunds can be delayed even though the attacker eventually receives their deposit amounts back. On chains with small withdrawal caps, this can turn deposit-contract spam into sustained validator-exit liveness pressure.

## Root Cause

Rejected-deposit refunds are admitted directly into the shared withdrawal queue before valid-deposit admission accounting and without a separate refund budget or priority rule.

## Code

- Rejected deposits with parseable withdrawal credentials create a `WithdrawalRequest` immediately: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1539, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1573.
[...]

PR #276, opened 2026-05-28
fix: separate refund withdrawal scheduling + add invalid deposit tax

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/275.

Addresses https://github.com/SeismicSystems/summit/issues/226 and https://github.com/SeismicSystems/summit/issues/234.

Changes:
- Separates deposit-refund withdrawals from validator withdrawals in scheduling and emission. Refunds now use the refund schedule while validator exits keep priority under their own withdrawal budget, preventing invalid-deposit refunds from consuming validator-exit capacity.
- Introduces a protocol parameter InvalidDepositTax. Invalid deposit refunds are now split into a net refund to the depositor withdrawal credentials and a tax amount to the configured treasury address. This removes a ddos vector, where people can spam the network with invalid deposit requests.
- Adds validation to the withdrawal queue invariants on decode.


sebastian-osec:
This looks directionally correct for refund isolation and the withdrawal-queue decode invariants: validator withdrawals and deposit-refund withdrawals are now separated by kind, real exits keep their own capacity, process-time top-up refunds use synthetic refund keys, and decode now rejects the main inconsistent queue states.

I have two questions:

1. `max_withdrawals_per_epoch` is now applied once to validator withdrawals and once to refund withdrawals, so a configured cap of N can emit up to 2N withdrawals in one terminal block. Is that intended as a per-kind cap? If the parameter is meant to remain a total per-epoch withdrawal cap, this seems like it needs either a separate refund cap or total-cap accounting with validator priority.

2. Because this changes the persisted/checkpoint encoding of `ConsensusState` / `PendingWithdrawal` / `WithdrawalQueue`, this would not support upgrading nodes from a pre-PR persisted state/checkpoint unless there is a migration or explicit state reset. Is that expected for this fix?

matthias-wright:
1. Good point. I think this is okay for now. We could add separate limits for the two withdrawal types, but I'm trying to avoid to many merge conflicts.
2. Yes, that is expected.

sebastian-osec:
Mostly LGTM from me after the later commits. I think the core starvation issue is addressed: validator withdrawals and deposit-refund withdrawals now have separate schedules, validator withdrawals are selected first, and refunds no longer consume validator-exit capacity.

One optional clarification: `get_withdrawals_for_epoch_with_limits` now supports separate validator/refund limits, but the finalizer currently passes `max_withdrawals_per_epoch` for both. So if `max_withdrawals_per_epoch = N`, the effective behavior is up to `N` validator withdrawals plus up to `N` refund withdrawals in the same terminal block.

Is that intended as the new per-kind meaning of `max_withdrawals_per_epoch`? If yes, it may be worth updating the docs to make that explicit. If the intent is to keep `max_withdrawals_per_epoch` as a total cap, then we may want either a separate refund config or total-cap accounting with validator withdrawals taking priority.

Small doc nit: `docs/deposits-and-withdrawals.md` still says refund withdrawals merge into an existing pending validator withdrawal, but the new queue rejects cross-kind merges and keeps refunds separate.

matthias-wright:
Update(5430f89):
- Make max_withdrawals_per_epoch a total cap with exit priority

sebastian-osec:
Looks good to me.

Only small cleanup nits:
- `finalizer/src/actor.rs` still has a stale comment saying “per-kind withdrawal budgets”.
- `docs/ssz-merklization.md` still describes `max_withdrawals_per_epoch` as “Max validator withdrawals per epoch”; it should say total withdrawals per epoch.

Other than that, I think the core issue is fixed.

Issue #242, opened 2026-05-22
[Security][tentative] Medium (correctness): DynamicEpocher Makes State Snapshots Porous


evonide:
## Context

`ConsensusState` is the shared view of validator accounts, epoch position, and protocol parameters used by finalizer, sync, checkpoint, and fork-handling paths. Summit clones this state when it needs a snapshot for an alternate fork or actor-local decision, so cloned states are expected to preserve the rules in force at the time of capture.

`DynamicEpocher` is the component inside that state that maps heights to epoch boundaries and supports epoch-length changes. Snapshot isolation depends on each cloned consensus state retaining the epoch schedule that belonged to that snapshot.

## Claim

`ConsensusState` clones are not isolated snapshots of the epoch schedule because `DynamicEpocher` is backed by shared mutable state. A later schedule mutation through one clone can change how another retained clone interprets heights.

An honest epoch-length protocol update or epoch transition can occur after a `ConsensusState` has been cloned for a fork, checkpoint, sync decision, or actor-local snapshot; because the clone shares the same mutable `DynamicEpocher`, later schedule changes leak into the old snapshot and make it interpret heights under rules that were not in force when it was captured.

## Flow

The path is reachable when a consensus state is cloned, retained for a fork or actor, and then the shared `DynamicEpocher` is advanced or updated before the old snapshot is reused.

## Impact

Old fork snapshots or actors can observe a mutated epoch schedule that did not exist when their state was cloned. That can change whether a height is viewed as penultimate, last, current, or stale, affecting verification, sync delivery, checkpoint construction, and epoch transition behavior.

## Root Cause

`DynamicEpocher` clone semantics preserve shared mutable state instead of producing an immutable schedule snapshot tied to the cloned `ConsensusState`.

## Code

- `DynamicEpocher` uses shared mutable state: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/dynamic_epocher.rs#L29.
- Finalizer clones consensus state into fork state: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L769.
- Fork snapshots are stored for later use: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L816.
- Protocol parameter changes update epoch length through the shared epocher: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L712.
[...]

PR #277, opened 2026-05-28
fix: dynamic epocher snapshot isolation

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/242.

Changes:
  - Add isolated DynamicEpocher snapshots with copy-on-write schedule storage.
  - Keep actor live handles synchronized by publishing promoted fork schedules back to the canonical shared epocher.
  - Add regression and edge tests for snapshot isolation, live-handle propagation, and post-replace mutation isolation.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -218,6 +218,8 @@ impl<
             gauge
         };
 
+        let shared_state = state.clone_with_shared_epocher();
+
         (
             Self {
                 context: ContextCell::new(context),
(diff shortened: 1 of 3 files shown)


evonide:
LGTM! This PR makes normal consensus-state clones use isolated `DynamicEpocher` snapshots while keeping explicit live-handle sharing only for finalizer wiring and fork promotion so old snapshots cannot observe later epoch-rule mutations.

Issue #245, opened 2026-05-22
[Security][tentative] Medium (attackable): Finalizer Leaves Losing-Fork Height Waiters Pending


evonide:
## Context

The finalizer lets other components wait until a particular block digest has executed at a target height, using `NotifyAtHeight` waiters to resume proposal or verification work once the requested block is available. Fork handling can temporarily track competing digests at the same height until one branch becomes canonical.

When finalization selects one digest, the finalizer should resolve both the winning waiter and any waiters whose requested digest can no longer become canonical at that height. The disrupted flow is fork cleanup: losing fork/orphan state is pruned, but the waiter map is a separate structure that also needs completion.

## Claim

`NotifyAtHeight` waiters are keyed by exact `(height, block_digest)`, but finalization only completes the winning digest. When a different block finalizes at that height, fork/orphan cleanup prunes the losing state without completing or removing waiters for the losing digests.

A malicious block producer or peer cannot call the finalizer mailbox directly, but can indirectly trigger application proposal or verification code to register a `NotifyAtHeight` waiter for a future parent digest. If that digest is not yet executed in a fork state and a different digest later finalizes at the same height, the losing-digest waiter can remain pending after fork cleanup.

## Flow

The stale-waiter path requires a future-height waiter: when a request arrives after canonical height has already passed the target, or at the current canonical height with a mismatched digest, the finalizer immediately returns `false`. The reachable start condition is therefore a proposal or verification task that has already registered a waiter for height `H` and digest `D` while `H` is still ahead of the canonical state and before a conflicting digest at `H` finalizes.

## Impact

Proposal or verification tasks waiting on a losing fork can remain pending until their receiver is dropped. The finalizer also retains sender entries for finalized-away digests. Under repeated fork/orphan races, this can accumulate stale waiters and convert ordinary fork resolution into task stalls.

## Root Cause

Height waiters are keyed by exact digest, but finalization cleanup only notifies the winning digest and prunes fork/orphan state without completing waiters for losing digests at finalized heights.

## Code

- Future height notifications are stored by exact `(height, block_digest)`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L361.
[...]

PR #278, opened 2026-05-28
fix: stale NotifyAtHeight waiters when a conflicting digest finalizes

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/245.

Changes:
  - Split notifier handling for notarized execution vs finalized canonical execution.
  - Keep notarized fork execution scoped to exact (height, digest) waiters.
  - On finalization, resolve all waiters at finalized heights: canonical digest gets true, losing/stale digests get false, future waiters remain pending.
  - Add fork-handling regression and edge tests for losing, duplicate, dropped, stale, and future waiters.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -573,7 +573,7 @@ impl<
         }
 
         let new_height = block.height();
-        self.height_notify_up_to(new_height, block_digest);
+        self.height_notify_finalized_up_to(new_height, block_digest);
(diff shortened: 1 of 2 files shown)


evonide:
LGTM! This PR resolves stale `NotifyAtHeight` waiters on finalization by returning true only for the canonical finalized digest and false for losing or outdated digests while keeping future waiters pending.

Issue #212, opened 2026-05-22
[Security] Critical (correctness): Voluntary exits can break finalization quorum


evonide:
## Context

At epoch transitions, Summit derives the next active validator set and asks the orchestrator to start Commonware Simplex for that epoch. Consensus for a live network requires a sufficiently large active committee to keep proposing, voting, and finalizing blocks.

Validator exits are normal lifecycle operations. Summit therefore needs to ensure that applying accepted exits cannot leave the next epoch without enough live voting power to continue finalization.

## Claim

A normal sequence of authorized validator exits can reduce Summit's active validator set below the threshold needed for continued finalization, because Summit applies removals at the epoch transition without enforcing a minimum live-committee or quorum-preservation invariant.

## Flow

1. Active validators submit otherwise valid withdrawal or exit requests.
2. Finalized-block processing records those validators in `removed_validators`.
3. At the next epoch transition, Summit marks the removed validator accounts inactive.
4. The finalizer derives the next active validator list from accounts still marked `Active`.
5. Summit sends that reduced validator list to the orchestrator for the next Commonware Simplex epoch.
6. If the remaining active set is below the consensus quorum needed to make progress, the chain can stop finalizing future blocks.

## Impact

Applying exits that leave too few live validators prevents the network from producing and finalizing later blocks. Honest nodes may keep running, but the chain can stop making finalized progress once the remaining active set is below the quorum needed by consensus.

This is a consensus liveness failure: recovery requires operators or governance to restore enough active validators, reject or defer unsafe exits, or otherwise repair the validator set before finalization can resume.

## Root Cause

Validator-removal logic is not paired with a global minimum live-committee check at epoch transition. Summit updates accounts independently, rebuilds the active validator set, and hands that result to the next consensus epoch without checking that the resulting committee can still finalize.

## Code

- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1668 records active-validator withdrawals for removal.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1196 applies `removed_validators` at epoch transition.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L652 derives active validators only from accounts with `ValidatorStatus::Active`.
- https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L676 sends the next active validator list to the orchestrator.
[...]

PR #279, opened 2026-05-28
fix: add minimum validator protocol param

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/212.

Changes:
  - Added MinimumValidatorCount to protocol params, genesis, consensus state, SSZ hashing/proofs, and example genesis.
  - Track pending active-validator exits during an epoch.
  - Reject or skip exits that would reduce the active validator set below the configured minimum.
  - Reset the pending-exit counter at epoch transition.
  - Added regression tests for normal full exits and stake-bound force removals hitting the minimum-validator floor.


evonide:
LGTM. The validator floor logic looks sound and fixes the original active-set drain issue.

Two small things to consider.

1) Since the chain is not live yet the new consensus state fields seem fine but old (devnet) checkpoints / DB snapshots written before this PR may not decode under the new state format meaning they would need either a reset or a small migration if compatibility matters.

2) A malicious validator could submit duplicate withdrawals on the last block of an epoch. Since up to 16 withdrawal requests can be emitted per block, duplicates could consume the exit counter and cause later valid withdrawals in that same block to be skipped. This is happening because last block withdrawals are deferred via `pending_execution_requests` and not marked as `has_pending_withdrawal = true` resulting in the `pending_active_validator_exits` no being incremented for last block withdrawal requests.

One possible simplification is to track pending active exits by validator pubkey instead of a counter or to correctly mark the validator as pending before deferring the last-block withdrawal.

matthias-wright:
Update (09427cd):
- Track deferred exits in a per-pass pubkey set and skip duplicates before counting or re-queuing.
- Add a regression test covering [A, A, B].

evonide:
LGTM please recheck after rebase to make sure the fix survived unchanged.

Issue #254, opened 2026-05-22
[Security][tentative] Medium (correctness): Proposer Can Build Too-Far-Future Timestamp From Parent


evonide:
## Context

The application proposer chooses timestamps for newly built Summit blocks, and peer validators independently enforce timestamp rules before voting. The normal invariant is that an honest proposer should only emit blocks that satisfy the same monotonicity and future-window checks used by verifiers.

Summit must balance two timestamp constraints: each child must be later than its parent, and it must not be too far ahead of the verifier's local clock. The finding concerns the proposer-side path where preserving parent-child monotonicity can produce a timestamp that still violates the verifier future allowance.

## Claim

The proposer enforces parent-child timestamp monotonicity without enforcing the verifier's future-timestamp bound. When local time is not greater than the parent timestamp, it emits `parent.timestamp() + 1` even if that value is still outside the configured future allowance that honest verifiers apply.

After clock skew or an accepted parent near the future-timestamp limit, an honest proposer can select `parent.timestamp() + 1` to preserve monotonicity without waiting for peers' future-window predicate; peers then reject the otherwise well-formed child as too far in the future, causing missed proposal or notarization progress.

## Flow

This requires local wall clock to be at or behind the parent timestamp, such as after clock skew or after accepting a parent near the future-timestamp limit. The proposer then chooses `parent.timestamp() + 1` without waiting.

## Impact

After clock skew or a previously accepted future-timestamp parent, an honest proposer can create a block that is monotonic relative to the parent but still too far ahead for peers. Honest validators may reject or fail to notarize the block, causing missed slots or temporary consensus stalls until clocks catch up.

## Root Cause

Proposal timestamp selection enforces monotonicity relative to the parent but does not enforce the same future-window predicate used by verifiers.

## Code

- Proposal timestamp selection bumps to `parent_block.timestamp() + 1` when local time is behind the parent: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L480, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L482.
- Verification rejects non-increasing timestamps: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L630.
- Verification also rejects timestamps above `now_millis + allowed_timestamp_future_ms`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L638.
[...]

PR #280, opened 2026-05-29
fix: wait for proposal timestamps to enter future window

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/254.

  Fixes proposer timestamp selection when the parent timestamp is near the verifier future-window limit.

  Changes:
  - Wait before building a payload until `parent.timestamp + 1` is acceptable under `now + allowed_timestamp_future_ms`.
  - Keep child timestamps monotonic by selecting `max(now, parent.timestamp + 1)` after the wait.
  - Add focused tests for wait and no-wait timestamp selection.
  - Add unit tests.

### DIFF
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -54,6 +54,15 @@ fn select_proposal_timestamp(now_millis: u64, min_child_timestamp: u64) -> u64 {
     now_millis.max(min_child_timestamp)
 }
 
+/// Whether the wait required to bring the monotonic child timestamp into the
+/// verifier future window exceeds the leader window. When true the block could
+/// not be notarized before the leader rotates anyway, so the proposal should be
+/// abandoned rather than waited out (which would also tie up the application
+/// actor, delaying verify/certify handling).
+fn proposal_wait_exceeds_leader_window(wait: Duration, leader_timeout: Duration) -> bool {
+    wait > leader_timeout
+}
+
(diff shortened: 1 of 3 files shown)


evonide:
The timestamp mismatch looks fixed. The proposer now waits until the required child timestamp is inside the verifier’s accepted future window before building the payload.

One follow-up concern is that the wait happens inside `handle_proposal` meaning a longer wait could keep the application actor from promptly handling other things like verify or certify messages. Is the assumption that this wait is always short enough in practice or should the wait be moved outside the actor path?

matthias-wright:
That's a good point.
What if we check if the initial waiting time returned by `proposal_timestamp_wait` is larger than `leader_timeout_ms`, and if so, we abort immediately?

sebastian-osec:
Yes, I think that makes sense. If the initial `proposal_timestamp_wait` is greater than the leader timeout, this proposal cannot become valid within the current leader window anyway, so aborting immediately avoids tying up the application actor.

The important part is that this should abort proposal creation, not bypass the timestamp wait. We should still only build once `parent.timestamp() + 1` is inside the verifier future window.

matthias-wright:
Update(c5dd562):
- Abandon proposal when timestamp wait exceeds leader timeout

sebastian-osec:
Thanks. Looks good to me.

Issue #240, opened 2026-05-22
[Security][tentative] Medium (correctness): Checkpoint RPC Can Exceed Response Cap


evonide:
## Context

Checkpoint RPC methods export checkpoint state and associated block/finalized-header artifacts to clients that need onboarding or recovery data. Those artifacts can grow with validator and pending-state size, while the JSON-RPC server enforces a fixed maximum size on serialized responses.

## Claim

Checkpoint RPC responses serialize the full checkpoint, last block, and finalized header as JSON `Vec<u8>` fields under a fixed 10 MiB response cap. The broken invariant is that `getCheckpoint` and `getLatestCheckpoint` can fail to export otherwise valid local checkpoints once encoded artifacts exceed that cap.

Normal validator-set or pending-state growth can produce already-valid checkpoint artifacts whose JSON encoding exceeds the 10 MiB response cap. Any client that can reach the node's RPC port and calls `getCheckpoint` or `getLatestCheckpoint` then receives a capped-response failure instead of the artifact, but the client does not need to supply or corrupt checkpoint bytes.

## Flow

The path requires a checkpoint whose serialized checkpoint, last-block, and finalized-header byte arrays exceed the fixed JSON-RPC response limit after JSON encoding. Public network reachability depends on operator exposure of the RPC port; when exposed, the RPC server binds `0.0.0.0` and these read methods are not gated by the permissioned pause/unpause API.

## Impact

For a sufficiently large validator set or pending-state footprint, checkpoint RPC export fails for every remote client even though the node has a valid checkpoint locally. This breaks checkpoint-based onboarding and recovery, forcing replay from genesis or manual artifact transfer.

## Root Cause

Checkpoint artifacts are exposed as JSON `Vec<u8>` byte arrays through capped JSON-RPC responses instead of a binary, compressed, chunked, or separately bounded artifact transport.

## Code

- RPC responses are capped at `10 * 1024 * 1024`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/lib.rs#L52.
- The RPC server listens on `0.0.0.0:{port}`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/lib.rs#L59.
[...]

PR #281, opened 2026-05-29
chore: make RPC response limits configurable via CLI

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/240.

Changes:
  - Added --rpc-max-request-body-size.
  - Added --rpc-max-response-body-size.
  - Introduced shared default DEFAULT_RPC_BODY_LIMIT_BYTES = 50 MiB.
  - Applied configured limits to public/admin RPC and genesis RPC startup.


sebastian-osec:
This looks good to me as a practical fix for the hardcoded checkpoint RPC response cap. The limit is now configurable and the default is raised, though larger future checkpoints may still require operator tuning or a chunked/binary transport later.

Issue #241, opened 2026-05-22
[Security][tentative] Medium (attackable): Deposit Signatures Are Not Chain Bound


evonide:
## Context

Deposits authorize a node key and consensus key to join Summit validator state with specified withdrawal credentials and amount. Because the same software version can run on distinct deployments, the signed deposit domain needs to identify the intended chain or deployment boundary, not just the message shape.

## Claim

Summit deposit authorization signs a domain derived only from the static protocol version. The genesis namespace exists as chain-specific replay protection for the node and network, but deposit signing and verification do not include the namespace, genesis hash, EL chain ID, or deposit contract binding.

A malicious depositor or relay that obtains a valid deposit signature set for a node/consensus key pair on one Summit deployment can submit the same signed deposit fields to another deployment using the same protocol version, provided the target chain's normal deposit prechecks allow that account state and amount. Once those local state checks pass, verification rebuilds a protocol-version-only domain and accepts the replayed authorization without proving the intended chain.

## Flow

The replay requires the same deposit message fields and protocol version on both networks, no chain-specific value in the signed deposit domain, and target-chain state in which the validator is not blocked by an existing pending deposit or withdrawal and the resulting balance remains within the configured stake bounds.

## Impact

A valid deposit signature set from one Summit network can authorize the same node and consensus keys on another network with the same protocol version. This is distinct from the duplicated protocol-version-constant issue: even if the constant remains consistent, the domain is still not chain-specific.

## Root Cause

The deposit signature domain is derived from `PROTOCOL_VERSION` only, and both signing and verification use an empty crypto namespace.

## Code

- `PROTOCOL_VERSION` is a static constant: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/lib.rs#L42.
- Genesis has a `namespace` intended to prevent replay attacks: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L40.
- Node config carries that namespace: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/config.rs#L53, https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/config.rs#L95.
- RPC deposit signing hashes only `PROTOCOL_VERSION` into the domain: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L250.
[...]

PR #282, opened 2026-05-29
fix: bind deposit signatures to genesis hash

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/241.

Changes:
  - Add summit-deposit-v1 || PROTOCOL_VERSION || genesis_hash deposit domain.
  - Use the domain for RPC/admin signing and finalizer verification.
  - Update local deposit tooling and tests/helpers.
  - Add regression coverage for cross-genesis signature rejection.
  
  
Note: once this is merged, we have to update seismic-viem and the Python client.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -11,7 +11,7 @@ use commonware_consensus::simplex::scheme::bls12381_multisig;
 use commonware_consensus::simplex::types::Finalization;
 use commonware_consensus::types::Epoch;
 use commonware_cryptography::bls12381::primitives::variant::Variant;
-use commonware_cryptography::{Digestible, Hasher, Sha256, Signer, Verifier as _, bls12381};
+use commonware_cryptography::{Digestible, Signer, Verifier as _, bls12381};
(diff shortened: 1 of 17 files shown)


sebastian-osec:
This looks good to me. The deposit signature domain is now derived from a tagged protocol-version plus the genesis hash, and both the RPC/admin signing path and finalizer verification path use the same domain. Since `DepositRequest::as_message` includes that domain in the signed message, signatures from one genesis hash should no longer replay on another, and the cross-genesis rejection test covers the intended failure mode.

This binds deposit signatures to the EL genesis hash, which fixes cross-chain replay when deployments have different EL genesis hashes. One question: is EL genesis hash the full intended Summit deployment boundary? If two Summit networks can share the same EL genesis hash but differ by Summit namespace/config/deposit-contract context, then signatures would still replay across those deployments unless the domain also includes that Summit-specific namespace or contract binding.

matthias-wright:
Update(00f39ff):
- Deposit domain now also binds the Summit namespace

sebastian-osec:
Thanks, the latest update binding the deposit domain to both the EL genesis hash and Summit namespace addresses my question. I rechecked the signing and verification paths, and both now use the same deployment-bound domain. This looks good to me.

Issue #250, opened 2026-05-22
[Security][tentative] Medium (correctness): Observer Mode Still Signs Consensus


evonide:
## Context

Observer mode is meant to run a node that follows the chain and serves RPC without proposing or voting. Startup still shares the normal validator keystore path with the engine, so the important boundary is whether observer mode removes consensus signing material before consensus components initialize.

## Claim

`--observer` only changes the P2P transport identity and still passes the original validator keystore into the engine. The engine can then build a normal consensus signer from the validator BLS key, so observer mode does not actually fence proposal or vote signing.

An operator intending to run a passive observer can start `--observer` with an existing validator keystore, and because startup only changes the P2P transport key while still forwarding the original keystore, the process still loads the validator BLS key and can sign consensus messages for the same validator. This is an operator/startup misconfiguration path, not a remote peer path.

## Flow

The path is reachable when observer mode is configured with the validator keystore. The observer flag changes the P2P transport key but does not remove or gate the BLS consensus key before engine startup; an observer started with a non-validator consensus key would fall back to verifier mode instead of signing.

## Impact

If the observer's local state diverges from the main validator, the two processes can produce conflicting proposals or votes, causing equivocation risk, liveness loss, or confusing validator accountability.

## Root Cause

Observer mode is implemented only as a transport-identity change; the consensus signing key remains available to engine, scheme, finalizer, and application paths.

## Code

- CLI docs say observer mode is RPC-only and does not propose or vote: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L154.
- Observer startup derives a child signer for P2P transport: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L375.
- The unchanged validator `key_store` is still forwarded into engine config: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L647.
- `Engine::new` extracts the validator BLS private scalar from that keystore: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L133.
- `SummitSchemeProvider` creates a signing scheme only when that private key is in the participant set, and otherwise falls back to verifier mode: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/scheme.rs#L122.
- Finalizer startup enters the current epoch with the active validator set: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L253.
[...]

PR #283, opened 2026-05-29
fix: force verifier-only schemes in observer mode

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/250

Changes:
  - Force observer mode to use verifier-only consensus schemes, even when started from a validator keystore.
  - Add tests for verifier-only signing behavior and observer catch-up with validator BLS material.


sebastian-osec:
This looks good to me.

Issue #249, opened 2026-05-22
[Security][tentative] Medium (unknown): Losing-Fork Descendants Remain Live for Aux Data


evonide:
## Context

The finalizer keeps fork states so notarized non-canonical branches can be evaluated and used for aux data while consensus decides which branch finalizes. Once a conflicting branch finalizes, states that no longer descend from the finalized chain should no longer be usable as live parents.

Aux-data requests let the application build or verify blocks against a parent state identified by height and digest. The intended invariant is that retained fork state remains compatible with the finalized ancestry before it can serve as a parent for further work.

## Claim

When a fork loses at a finalized height, the finalizer prunes fork states only at or below that height. Descendants above the finalized height can remain in `fork_states`, and later `NotifyAtHeight` or `GetAuxData` calls treat those descendants as valid parents without rechecking ancestry against the canonical finalized chain.

A finalizer-received sequence where a fork descendant is notarized before a conflicting branch finalizes at the ancestor height can prune only the losing ancestor, leaving the descendant in `fork_states` so later proposer or verifier aux-data requests still treat the stale branch as live. The claim is limited to the finalizer state-machine behavior once such messages are delivered; it does not assume honest Commonware will naturally produce the conflicting notarization/finalization order.

## Flow

Reachable at the finalizer API/state-machine layer when a fork has a retained notarized descendant above a height where a conflicting branch finalizes, for example `A1 -> A2` followed by finalization of conflicting `B1`. The required prerequisite is delivery or construction of that notarized-descendant-then-conflicting-finalization sequence.

## Impact

If fork `A1 -> A2` is notarized and a conflicting `B1` finalizes at height 1, `A1` is pruned but `A2` can remain live. Honest nodes can then build or verify children of `A2` using valid-looking aux data even though that branch no longer extends the finalized chain. This can prolong dead forks, waste views, and create a path for stale fork state to be reused as canonical if later finalized handling sees that digest.

## Root Cause

Fork-state pruning is height-based rather than ancestry-based, and aux-data lookup accepts retained fork states by height/digest without checking compatibility with the finalized chain.

## Code

- Notarized fork blocks are executed into full fork states: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L804.
- Those fork states are stored by height and digest: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L816.
[...]

PR #284, opened 2026-05-29
fix: prune stale fork descendants on finalization

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/249.

Changes:
  - Track each fork state’s parent digest.
  - On finalization, keep only retained forks that descend from the finalized block.
  - Tombstone pruned fork digests so later waiters fail immediately.
  - Add a regression for A1 -> A2 followed by conflicting B1 finalization.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -914,7 +914,8 @@ impl<
         // execute a non-canonical block onto canonical state.
         let canonical_height = self.canonical_state.get_latest_height();
         let canonical_head = self.canonical_state.get_head_digest();
-        if height == canonical_height + 1 && block.parent() != canonical_head {
+
+        if height != canonical_height + 1 || block.parent() != canonical_head {
             error!(
                 target: "critical",
                 height,
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -251,6 +251,7 @@ fn queue_deposit_refund(
 #[derive(Clone, Debug)]
 struct ForkState {
     block_digest: Digest,
+    parent_digest: Digest,
     consensus_state: ConsensusState,
 }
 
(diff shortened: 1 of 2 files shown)


sebastian-osec:
This fixes the stale notarized fork-state / aux-data path: after a conflicting ancestor finalizes, descendants that no longer extend the finalized block are pruned/tombstoned, NotifyAtHeight fails, and GetAuxData no longer serves the dead descendant.

One remaining path I think is worth handling: if that pruned descendant later arrives via FinalizedBlock, handle_finalized_block does not appear to check the tombstone or verify that the block extends the current canonical head before directly executing it. Should finalized-block handling reject tombstoned or non-canonical-parent blocks as well?

matthias-wright:
Update(a98b90d):
- Added check that a finalized block extends the canonical chain.

sebastian-osec:
Thanks, this handles the immediate losing-descendant case where the incoming finalized block is exactly `canonical_height + 1`.

I think there is still a gap if finalized input is ever non-sequential. The new guard only runs when `height == canonical_height + 1`; if a deeper pruned descendant later reaches `handle_finalized_block`, e.g. `A1 -> A2 -> A3`, then conflicting `B1` finalizes, and `A3` arrives as `FinalizedBlock` while canonical height is still `1`, the parent check is skipped because `3 != 1 + 1`. There is also no `dead_fork_digests` check in `handle_finalized_block`, so the dead descendant can still fall through to fork-state lookup / direct execution.

Would it be safer to make finalized handling enforce sequencing before execution, e.g. reject or buffer `height != canonical_height + 1`, and separately reject `(height, digest)` entries present in `dead_fork_digests`? A regression for `A1 -> A2 -> A3`, conflicting `B1` finalization, then `A3` via `FinalizedBlock` would cover this remaining path.

matthias-wright:
Update(a2b2dc9);
- Changes the finalization guard to `if height != canonical_height + 1 || block.parent() != canonical_head`.
- The syncer sends finalized blocks in order, so if a finalized block arrived out of order, the finalizer will abort with an error..

Update(3b0f1a5):
- Adds test for the scenario: notarize A1 -> A2 -> A3, then conflicting B1 finalizes, and A3 arrives as FinalizedBlock while canonical height is still 1

I think we don't need to check `dead_fork_digests` in `handle_finalized_block` anymore with these changes:
- The guard halts unless height == canonical_height + 1 && block.parent() == canonical_head.
- A digest is in dead_fork_digests because prune_fork_states_not_descending_from found its parent_digest was not on the canonical chain (it didn't descend from the finalized head).
- canonical_head only moves forward to newly finalized blocks — it never rewinds to a pruned fork's parent. So a dead fork's parent can never become the canonical head later.
- Therefore a dead-fork digest can never satisfy parent == canonical_head at height == canonical+1.

sebastian-osec:
Looks good to me. I agree that with the new strict finalized-block guard, we do not need a separate `dead_fork_digests` check in `handle_finalized_block`.

Issue #238, opened 2026-05-22
[Security][tentative] Medium (correctness): Checkpoint Restart Drops Exiting Validators Early


evonide:
## Context

Summit keeps the current-epoch signing committee intact through the epoch boundary, even when some validators have already submitted exits for future membership changes. Checkpoint restart reconstructs runtime validator and discovery sets from checkpoint consensus state, so it must preserve the difference between current-epoch signers and future-epoch active members.

## Claim

Finalizer startup reconstructs the current epoch committee using only `Active` accounts, while checkpoint verification still treats `SubmittedExitRequest` validators as current-epoch signers. The broken invariant is that a checkpoint restart can drop exiting validators from the current committee before the epoch boundary.

An operator or recovery automation imports or restarts from a penultimate checkpoint that legitimately contains `SubmittedExitRequest` validators; startup rebuilds the current-epoch committee from active-only accounts, so the recovered node excludes validators that live peers still require through the epoch's final block.

## Flow

The path is reachable when a withdrawal is processed before checkpoint creation near the epoch boundary, the resulting checkpoint is imported or used for restart, and startup rebuilds the current epoch from checkpoint state rather than from the original epoch committee.

## Impact

Live nodes keep the original epoch committee through the last block, while a checkpoint-restored node can reject valid finalizations from live peers, fail to contribute to the last block, or use an inconsistent verifier set during recovery. Startup also stops tracking the still-current exiting validator and observer keys in the P2P oracle. If many nodes restart from the same checkpoint, the network can enter the final block with a smaller unauthorized committee and reduced discovery reachability for legitimate current-epoch participants.

## Root Cause

Startup conflates "active for future membership" with "still in the current epoch signing set" by calling `get_active_validators()` during current-epoch reconstruction.

## Code

- A non-last-block withdrawal marks an active validator `SubmittedExitRequest` and records it in `removed_validators`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1669.
- Execution requests are processed before checkpoint creation: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1391.
- The checkpoint is created at the penultimate block: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1423.
[...]

PR #285, opened 2026-05-29
fix: checkpoint restart current-epoch committee

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/238.

Changes:
  - Adds a current-epoch validator helper that includes Active and SubmittedExitRequest.
  - Uses that set for finalizer startup oracle tracking and initial orchestrator Enter.
  - Adds a regression test for checkpoint-like restart with a pending exit.


sebastian-osec:
This looks good to me.

Issue #261, opened 2026-05-22
[Security][tentative] Medium (attackable): Typed Finalized Headers Can Bypass Signed Field Binding


evonide:
## Context

Finalized headers carry consensus header fields together with a finalization certificate that should authenticate the header digest. Summit also exposes typed Rust structures and verifier helpers for checkpoint-chain validation, so those typed APIs form a trust boundary for integrations that operate on already-decoded values.

The intended invariant is that any finalized header accepted by the verifier has its public header fields bound to the certificate payload. This finding concerns the gap between decode-time validation and later typed-object use.

## Claim

SSZ decoding verifies that a `FinalizedHeader` certificate signs the header digest, but the public typed struct and constructor do not preserve that invariant. `verify_checkpoint_chain()` accepts already-constructed `FinalizedHeader` values and trusts their mutable header fields without rechecking that the certificate payload still equals the header digest.

A malicious checkpoint supplier targeting a downstream typed-API integration can provide or induce an already-constructed `FinalizedHeader` with a valid finalization certificate but mutated public header fields. `verify_checkpoint_chain()` then trusts those unsigned fields because the payload/header binding was enforced only during SSZ decode.

## Flow

The issue is reachable for typed API callers that bypass SSZ decode-time validation, including tests or integrations that construct `FinalizedHeader` directly or mutate its public fields after construction. It is not currently shown to be reachable through the stock startup checkpoint-file path, because that path decodes finalized headers with `FinalizedHeader::from_ssz_bytes` before verification.

## Impact

A caller that treats `verify_checkpoint_chain()` as the trust boundary for untrusted typed artifacts can pass a valid finalization certificate paired with mutated header fields. In particular, `checkpoint_hash` can be changed to match attacker-controlled checkpoint data while the certificate still verifies against the old digest. Startup SSZ import is partially protected by decode-time checks, but the typed verifier API can still break trustless checkpoint validation for downstream callers or tests.

## Root Cause

`FinalizedHeader` exposes mutable public fields and a constructor that does not bind `finalization.proposal.payload` to `header.digest`; the checkpoint verifier accepts typed values without recomputing that binding.

## Code

- `FinalizedHeader` exposes public `header` and `finalization` fields: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/header.rs#L327.
[...]

PR #286, opened 2026-05-29
fix: bind finalized-header fields to the signed certificate

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/170.

Addresses https://github.com/SeismicSystems/summit/issues/261.

Changes:
- Header: add computed_digest() (recompute from fields, ignoring the seedable OnceLock cache); use it for all binding checks.
- FinalizedHeader: make fields private; new() validates the payload/digest binding (returns FinalizedHeaderError), new_unchecked() for callers that already established it; add accessors.
- verify_checkpoint_chain: recompute and enforce finalization.proposal.payload == header digest per header, before trusting any field (PayloadDigestMismatch).
- Header: hand-write PartialEq/Eq to exclude the memoized digest cache so equality is field-based regardless of cache warmth.
- Add negative tests for the new() rejection and verifier rejection.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -903,9 +903,14 @@ impl<
             // Get participant count from the certificate signers
             let participant_count = finalization.certificate.signers.len();
 
-            // Store the finalized block header in the database
-            let finalized_header =
-                FinalizedHeader::new(block.header.clone(), finalization, participant_count);
+            // Store the finalized block header in the database. The binding
+            // between the block header and this finalization was just
+            // established by consensus (asserted above), so construct directly.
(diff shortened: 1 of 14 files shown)


evonide:
LGTM! This makes finalized-header fields private and requires the header fields to hash back to the signed finalization payload before checkpoint verification trusts fields like `checkpoint_hash`, `epoch` or validator deltas.

One small caveat: `new_unchecked` is still public, so future trusted callers could bypass the constructor check if they use it carelessly. That does not block this PR because decode and `verify_checkpoint_chain` still recheck the binding, but it is worth keeping `new_unchecked` limited to places that already proved the header/certificate pairing.

Issue #253, opened 2026-05-22
[Security][tentative] Medium (attackable): Pending Checkpoint Digest Is Not Validated


evonide:
## Context

Consensus state can carry a `pending_checkpoint` that will later be used to finalize a checkpoint artifact. The intended invariant is that each checkpoint's digest is the hash of its embedded data, because trustless recovery compares finalized checkpoint hashes against the checkpoint bytes.

## Claim

Decoded checkpoints carry `data` and `digest` as independent fields, and `ConsensusState.pending_checkpoint` decoding accepts the embedded digest without checking it matches the embedded data. The finalizer later uses that stored digest directly as the epoch-final block's checkpoint hash.

A malicious unchecked checkpoint artifact supplier or local decoded-state source can provide consensus state containing a `pending_checkpoint` whose stored digest differs from its data hash, and finalization later copies that unvalidated digest into the checkpoint hash so the exported checkpoint artifact fails trustless verification. A trustlessly verified checkpoint is not an arbitrary edit surface unless the signed checkpoint data itself already contains the malformed embedded checkpoint.

## Flow

The path is reachable through consensus-state decoding when `pending_checkpoint` is present. It requires malformed serialized state, an explicitly trusted checkpoint loaded without verification headers, or canonical signed checkpoint data that already contains `pending_checkpoint.data = X` and `pending_checkpoint.digest = Y` where `Y != sha256(X)`, followed by finalization that consumes the pending checkpoint. Startup verifies the outer checkpoint only when finalized headers are supplied.

## Impact

Honest proposers that start from or otherwise decode such malformed state can finalize a header committing `Y`, while the stored/exported checkpoint artifact contains `X`. Future trustless importers reject the checkpoint because the signed header hash does not match the checkpoint data hash, making that epoch's checkpoint unusable for recovery.

## Root Cause

The `Checkpoint` type stores a redundant digest but decode paths do not rederive and validate it before the finalizer trusts it as the canonical checkpoint hash.

## Code

- `Checkpoint::new` defines the intended data/digest invariant: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L26.
- `Checkpoint::from_ssz_bytes` decodes `data` and `digest` independently without validating their relationship: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L75.
[...]

PR #288, opened 2026-06-01
fix: reject pending checkpoints in decoded checkpoint artifacts

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/253

Changes:
- Rejects checkpoint artifacts that decode to consensus state with a non-None pending_checkpoint. This closes the checkpoint recovery path that could carry staged checkpoint data into finalized artifacts.
- Adds a regression test.

### DIFF
--- a/types/src/checkpoint.rs
+++ b/types/src/checkpoint.rs
@@ -86,9 +86,20 @@ impl Decode for Checkpoint {
         let data: Vec<u8> = decoder.decode_next()?;
         let digest_bytes: [u8; 32] = decoder.decode_next()?;
 
+        // Bind the redundant `digest` field to `data`: a decoded checkpoint must
+        // satisfy `digest == sha256(data)`.
+        let digest = Digest::from(digest_bytes);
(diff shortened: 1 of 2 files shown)


evonide:
This is a good improvement but seems narrower than #253. It rejects finalized checkpoint artifacts that contain `pending_checkpoint` which closes the main checkpoint import path.

However, the underlying invariant was that a `Checkpoint` should not be trusted with `digest != sha256(data)` and raw `Checkpoint::read_cfg` / `ConsensusState::read_cfg` still appear to accept an embedded pending checkpoint without validating that binding. Could we either validate the checkpoint data/digest when decoding/reading `Checkpoint` or recompute/check the pending checkpoint digest right before the finalizer consumes it?

matthias-wright:
Update(20d3847);
- Validate checkpoint digest against data on decode

sebastian-osec:
This fixes the underlying invariant. Looks good to me.

Issue #247, opened 2026-05-22
[Security][tentative] Medium (attackable): Genesis Simplex Timeouts Are Not Validated


evonide:
## Context

Summit uses Commonware Simplex as its consensus engine, and genesis carries the launch-time timeout values that shape leader, notarization, nullification, activity, and skip behavior. These values cross the startup trust boundary before the engine starts and then drive consensus scheduling.

The normal startup flow loads genesis, copies those timeout fields into node and engine configuration, and constructs the Simplex engine. Simplex requires nonzero and correctly ordered timeout values, so Summit's genesis validation is expected to reject configurations that cannot be used to initialize consensus.

## Claim

Genesis validation does not enforce the timeout invariants required by Commonware Simplex. Invalid genesis values are passed into Simplex engine construction, where the dependency enforces those invariants with hard assertions.

A malicious genesis distributor, compromised bootstrap service, compromised deployment pipeline, or first-boot genesis RPC sender supplies a genesis file with zero or misordered Simplex timeout values; Summit accepts the file during genesis loading, then passes the invalid values into Commonware engine construction where startup panics.

## Flow

The path is reachable during normal node startup using a malformed genesis configuration with zero timeout values or invalid timeout ordering. Checkpoint recovery does not introduce a separate timeout source: the node may load consensus state from a checkpoint, but the Simplex timeout configuration still comes from the genesis file loaded during startup.

## Impact

A malformed or malicious genesis file can deterministically panic every node during first epoch startup. The same bad genesis file can also break later checkpoint-recovery startup because the engine is still configured from genesis; checkpoint consensus state itself does not carry these timeout fields.

## Root Cause

Summit validates some genesis fields but omits the timeout invariants that the downstream Commonware Simplex config treats as mandatory assertions.

## Code

- Genesis exposes `leader_timeout_ms`, `notarization_timeout_ms`, `nullify_timeout_ms`, `activity_timeout_views`, and `skip_timeout_views`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L17.
- `Genesis::load_from_file` validates block count, timestamp future bounds, and treasury address, but not these timeout invariants: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L135.
[...]

Issue #224, opened 2026-05-22
[Security][tentative] High (attackable): Genesis Can Configure a Zero Withdrawal Cap


evonide:
## Context

Summit uses a per-epoch withdrawal cap to bound how many ready exits, refunds, and withdrawals are emitted in the terminal block for an epoch. The cap must be positive so the withdrawal queue can make forward progress while epoch-final auxiliary data remains bounded.

Genesis and checkpoint or consensus-state material can seed the same protocol parameter before normal block processing resumes. The intended invariant is that every ingress path creating consensus state preserves the runtime range checks, because the finalizer later treats the stored cap as authoritative.

## Claim

Genesis and decoded consensus state accept `max_withdrawals_per_epoch = 0` even though runtime protocol-parameter updates reject it. With a zero cap, the epoch-final withdrawal selector emits no withdrawals and leaves due withdrawals to be rescheduled.

A malicious genesis distributor, first-boot genesis RPC sender, or malicious checkpoint/state provider can supply consensus state with `max_withdrawals_per_epoch = 0`, and nodes that start or restore from that state continue accepting deposits and exits while the epoch withdrawal path emits nothing and rolls every due withdrawal forward.

## Flow

The path is reachable at genesis load, first-boot genesis RPC delivery, and operator-supplied checkpoint or decoded consensus-state ingress because those paths do not apply the runtime protocol-parameter range check. Exploitation requires startup or restore from a state with the zero cap; live blocks, P2P messages, steady-state RPC, Engine API payloads, deposits, and runtime protocol-parameter updates are not the claimed way to set this value after a valid launch.

## Impact

A chain launched or restored with `max_withdrawals_per_epoch = 0` can accept deposits and schedule refunds/exits, but never emit any withdrawal. Validator exits, stake-bound withdrawals, and deposit refunds roll forward indefinitely until a valid protocol update raises the cap. The supported impact is startup/bootstrap or restore-state acceptance, not a canonical live transition that forces an already-running honest chain into the zero cap.

## Root Cause

The protocol-parameter invariant is enforced for runtime updates but not at all ingress points that can create consensus state, leaving genesis and decoded state able to bypass the lower bound.

## Code

- Runtime protocol-param validation enforces a minimum of one withdrawal per epoch: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/protocol_params.rs#L137.
[...]

Issue #223, opened 2026-05-22
[Security][tentative] High (attackable): Genesis Max Withdrawals Cap Can Oversize Epoch-Final Blocks


evonide:
## Context

Summit stores deposit, exit, refund, and withdrawal limits as protocol parameters in consensus state. The withdrawal cap bounds how many ready withdrawals the terminal block for an epoch may carry, keeping epoch-final auxiliary data within the same limits expected by proposers, verifiers, the execution layer, and P2P propagation.

At startup, a genesis file seeds the initial consensus state before any runtime update path runs. The intended invariant is that launch-time protocol parameters satisfy the same bounds as later protocol-parameter updates, because epoch-final block construction treats the stored withdrawal cap as authoritative.

## Claim

Genesis loading does not validate `max_withdrawals_per_epoch` against the documented `1..=256` runtime-update range. An oversized genesis value is copied directly into consensus state and later controls terminal-block withdrawal selection.

A malicious or compromised genesis artifact supplier, deployment pipeline, or first-boot genesis RPC sender can launch a chain with an oversized `max_withdrawals_per_epoch`, and once enough withdrawals are ready, honest proposers and verifiers follow that unchecked genesis cap when constructing and validating epoch-final withdrawal payloads.

## Flow

The path is reachable only through startup or first-boot genesis provisioning: Summit either loads the configured local genesis file or waits for the first-boot genesis RPC to write one, then seeds consensus state from that file. The impact requires enough ready withdrawals to make the selected terminal-block list exceed expected block, EL, or P2P limits; no live block, RPC, P2P, deposit, or runtime protocol-parameter path has been shown to raise the cap after launch.

## Impact

A chain launched with a very large withdrawal cap can force honest proposers to build epoch-final blocks with arbitrarily many withdrawals once enough are ready. Honest proposers cannot safely truncate because verifiers compute the same oversized aux data and reject mismatches. The terminal block can become too large for EL/P2P limits or practically unpropagatable, stalling epoch finalization.

## Root Cause

Genesis validation does not apply the same `max_withdrawals_per_epoch` bounds as runtime protocol parameter updates.

## Code

- Docs say `max_withdrawals_per_epoch` is range-limited and controls the last block of each epoch: https://github.com/SeismicSystems/summit/blob/ed2c5c8/docs/deposits-and-withdrawals.md#L112.
[...]

Issue #221, opened 2026-05-22
[Security][tentative] High (attackable): Genesis Epoch Length Can Exceed Runtime Maximum


evonide:
## Context

Genesis configuration seeds Summit's initial protocol parameters before any runtime governance update can occur. The epoch length controls when checkpoint creation, final-block withdrawals, committee transitions, and queued protocol-parameter changes execute, so the launch value should satisfy the same bounds as later protocol-param updates.

## Claim

`Genesis::load_from_file` accepts any nonzero `blocks_per_epoch`, including values above the runtime `MAX_EPOCH_LENGTH`, and startup passes that unchecked value into `ConsensusState::new`.

A malicious first-boot genesis RPC caller can supply a genesis with `blocks_per_epoch` greater than the runtime maximum before startup loads the configured genesis file. The network can start and produce ordinary blocks, but all epoch-boundary mechanics are deferred until the oversized boundary, making withdrawals, validator activation/removal, checkpoints, and accepted governance changes unavailable until that boundary is reached; a trusted genesis author or operator can trigger the same bug as a lower-friction correctness variant.

## Flow

The path requires launch from a local or provisioned genesis file. Runtime `EpochLength` requests reject values above `MAX_EPOCH_LENGTH`, so the currently source-backed trigger is genesis configuration rather than a post-launch public input.

## Impact

An oversized launch epoch can turn epoch-boundary functionality into a practical liveness failure for validator lifecycle, checkpoint recovery, withdrawals, and governance updates. This is lower severity than a remote live-network attack because the start condition is trusted launch configuration.

## Root Cause

Genesis validation checks only `blocks_per_epoch == 0` and does not reuse the protocol-parameter lower and upper bounds for epoch length.

## Code

- Runtime bounds define `MAX_EPOCH_LENGTH = 1_814_400`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/protocol_params.rs#L7-L8.
- Genesis loading only rejects `blocks_per_epoch == 0`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/genesis.rs#L135-L155.
- Startup converts the genesis value to `NonZeroU64` and passes it into `ConsensusState::new`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L709-L735.
[...]

PR #289, opened 2026-06-01
fix: genesis validation

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/271.

Addresses https://github.com/SeismicSystems/summit/issues/247, https://github.com/SeismicSystems/summit/issues/224, https://github.com/SeismicSystems/summit/issues/223, https://github.com/SeismicSystems/summit/issues/221.

Adds validation for genesis entries.

Changes:
- Epoch length - genesis blocks_per_epoch must be MIN_EPOCH_LENGTH..=MAX_EPOCH_LENGTH.
- Max withdrawals/epoch - reject out-of-range max_withdrawals_per_epoch when decoding ConsensusState (covers checkpoint and raw-state load), matching genesis.
- Stake bounds - reject inverted min > max at genesis and on runtime stake updates, which would otherwise make the balance check unsatisfiable.
- Simplex timeouts - validate genesis timeout/view fields are nonzero and ordered.
- Max message size - genesis max_message_size_bytes must be 1 MiB..=1 GiB.

### DIFF
--- a/types/src/consensus_state.rs
+++ b/types/src/consensus_state.rs
@@ -1187,6 +1187,21 @@ impl Read for ConsensusState {
 
         let max_deposits_per_epoch = buf.try_get_u64().map_err(|_| Error::EndOfBuffer)?;
         let max_withdrawals_per_epoch = buf.try_get_u64().map_err(|_| Error::EndOfBuffer)?;
+        // Enforce the same lower/upper bound the runtime protocol-parameter update
+        // path applies (see ProtocolParam::read_cfg / try_from). Genesis and runtime
+        // updates already range-check this, so an out-of-range value here means a
+        // crafted checkpoint/state artifact or tampered blob. A zero cap would let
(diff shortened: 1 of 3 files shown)


evonide:
LGTM. This fixes the reviewed genesis/restored-state validation issues by rejecting unsafe imported config before it can become consensus state.

**One long-term maintenance point to consider**: the same invariants are now checked across runtime params, genesis validation, consensus-state decode, and pending-param application so it may be worth eventually centralizing the installed-protocol-config validation to avoid future drift. Otherwise, we can have one-offs with invariants going forwards.

matthias-wright:
Update(1b68dd9);
- Added centralized validation for protocol params

evonide:
LGTM! The centralized `ProtocolParam::validate` path addresses the earlier drift concern and the unsafe values are rejected before becoming consensus state.

Issue #222, opened 2026-05-22
[Security][tentative] High (attackable): Genesis File Presence Can Brick First-Boot Provisioning


evonide:
## Context

First-boot provisioning lets a node wait for genesis material before entering normal startup. The intended invariant is that the waiting path should close only after usable genesis content is available, because malformed or partial genesis data still needs operator or provisioning recovery.

Summit uses a configured genesis file path as the handoff between provisioning and startup. This finding concerns the case where path existence is treated as readiness before the file has been parsed, validated, or atomically installed.

## Claim

First-boot provisioning treats genesis path existence as readiness, so an empty, malformed, or partial file disables provisioning and bricks restart recovery until local repair.

For the critical path, a remote first-boot RPC caller reaches the unauthenticated genesis provisioning endpoint and writes empty, malformed, or partial genesis bytes to the configured path. Restart treats path existence as successful provisioning, disables the waiting RPC path, and crashes on every load attempt until local repair; the standalone operator, automation, filesystem-fault, or interrupted-write case is a lower-severity correctness trigger.

## Flow

This is reachable during first boot before genesis has been successfully provisioned. It requires the configured genesis path to exist with empty, malformed, or partially written contents, which can come from local automation, operator pre-creation, filesystem failure, or an interrupted direct write. Once a valid genesis file already exists, this sentinel bug is no longer the active start condition.

## Impact

An operator automation mistake, pre-created file, disk-full event, or interrupted genesis provisioning write can trap first boot until local repair. The node sees the file as provisioned, refuses the normal RPC provisioning flow, then fails while parsing the malformed file on every restart until the file is manually removed or replaced.

## Root Cause

Genesis availability is represented by path existence rather than successful parse/validation, and RPC provisioning writes directly to the target path without atomic staging.

## Code

- `has_file` returns true when the path exists, regardless of whether the content is nonempty or parse-valid: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L169.
- `check_sender` sends the startup signal and disables the one-shot sender based on that predicate: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L178.
- Startup waits for the signal, cancels the genesis RPC server, and then `expect`s `Genesis::load_from_file`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L260, https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L265.
[...]

PR #290, opened 2026-06-01
fix: validate genesis before treating path as provisioned

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/222.

Changes:
- acquire_genesis replaces has_file/check_sender and classifies the configured path: absent → run the provisioning RPC and wait; present + valid → load it (RPC never exposed); present + invalid → exit with a clear error instead of crash-looping or falling back to provisioning.
- sendGenesis now stages to a temp file in the same dir, validates it with Genesis::load_from_file, then atomically renames into place; invalid content is rejected (RpcError::InvalidGenesis) and cleaned up. The target path never holds partial/invalid genesis.
- Telemetry is initialized before genesis acquisition so first-boot provisioning (which blocks waiting on the RPC) and the RPC's own "listening" line are actually logged - previously the subscriber wasn't installed until after this point and those logs were dropped.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -186,26 +186,6 @@ impl Command {
         }
     }
 
-    fn has_file(path: &str) -> bool {
-        let path_buf = get_expanded_path(path).expect("Invalid filepath");
-        path_buf.exists()
-            || !std::fs::read_to_string(&path_buf)
-                .unwrap_or_default()
-                .trim()
-                .is_empty()
-    }
-
-    fn check_sender(path: String, tx: oneshot::Sender<()>) -> PathSender {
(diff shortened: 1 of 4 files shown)


evonide:
LGTM. This fixes the malformed/partial genesis provisioning issue by validating staged genesis content before installing it and by refusing to treat an invalid existing file as provisioned. One small test hardening idea would be adding a direct startup/acquire-genesis regression for an already-present invalid file but the core issue looks fixed.

matthias-wright:
Update(e1012fd);
- Cover already-present invalid genesis at startup

evonide:
LGTM!

Issue #259, opened 2026-05-22
[Security][tentative] Medium (correctness): SSZ State Root Omits Pending Execution Requests


evonide:
## Context

Summit's execution layer sends consensus-affecting requests such as deposits, withdrawals, and exits into finalizer state. Requests that must be deferred across a boundary are stored in `pending_execution_requests` and replayed later so future validator balances and statuses reflect earlier execution activity.

The SSZ state tree supplies the advertised state root used in finalized aux data and proof flows. For that root to bind consensus state, it must cover deferred execution work as well as immediately applied account and committee fields.

## Claim

The SSZ state root omits `pending_execution_requests` even though deferred deposits, withdrawals, and exits are serialized consensus state that affect future transitions.

An honest block transition buffers deferred execution work in `pending_execution_requests`; the serialized state then differs in deferred deposits, withdrawals, or exits while `SszStateTree` omits that field, so the advertised SSZ state root and proofs remain unchanged.

## Flow

The live start condition is narrower than any execution-request processing: `parse_execution_requests` first drains existing pending requests with `take_pending_execution_requests`, then later code can buffer a request such as a last-block active-validator withdrawal with `push_pending_execution_request` before `capture_state_root`. The prerequisite is a captured or rebuilt state that contains deferred request bytes; ordinary request replay that leaves the pending vector empty does not demonstrate this issue.

## Impact

The `parent_beacon_block_root` and generated SSZ proofs do not commit to all consensus-relevant state. Checkpoint/restart/proof consumers can agree on a state root while disagreeing about deferred execution requests that schedule future deposits, withdrawals, or exits. That breaks the root-binds-state property expected from finalized roots and checkpoint proofs.

## Root Cause

`pending_execution_requests` were added to serialized consensus state and transition logic but not added as a leaf/input to `SszStateTree` root construction.

## Code

- `ConsensusState` stores `pending_execution_requests`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L33.
- Live execution mutates pending execution requests: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L388, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1473, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1664.
[...]

PR #291, opened 2026-06-01
fix: bind pending_execution_requests into the SSZ state root

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/259.

Changes:
- Add `PENDING_EXECUTION_REQUESTS_ROOT` (leaf 22) to the top-level tree with a subtree rebuilt like the other collections; collection root is `mix_in_length(subtree.root(), count)`.
- Each request blob is hashed as an SSZ byte list (`hash_byte_list`: pack into 32-byte chunks, merkleize, mix in byte length).
- Thread through `SszStateTree::rebuild` and `ConsensusState::rebuild_ssz_tree`, and keep the subtree in sync on `push_pending_execution_request` / `take_pending_execution_requests` so `capture_state_root` reflects it.

### DIFF
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -36,7 +36,7 @@ The state tree is a two-level design: a fixed top-level tree containing scalar f
 
 ### Top-Level Tree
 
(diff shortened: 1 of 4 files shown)


sebastian-osec:
LGTM. This binds the omitted consensus-state field into the rebuilt and captured SSZ state root, so changing that field now changes the advertised root.

Issue #258, opened 2026-05-22
[Security][tentative] Medium (attackable): SSZ State Root Omits Pending Checkpoint


evonide:
## Context

Summit's finalizer prepares checkpoint data around epoch boundaries. At the penultimate block it stores a `pending_checkpoint`, and the last block carries the checkpoint hash that commits the boundary artifact into the finalized chain.

The intended invariant is that the advertised SSZ state root for checkpoint or proof consumers authenticates the consensus-state fields needed to verify the next boundary transition. `pending_checkpoint` is one of those fields because it determines the checkpoint digest that the boundary header is expected to carry.

## Claim

`pending_checkpoint` is serialized state used to derive the next boundary `checkpoint_hash`, but it is omitted from the SSZ state root used for `parent_beacon_block_root`.

A malicious checkpoint producer or state artifact supplier can persist or serve consensus state with a serialized `pending_checkpoint` that differs from the value expected by a consumer while advertising the same SSZ state root, because `pending_checkpoint` is omitted from the root tree and the root alone cannot bind the next boundary checkpoint digest.

## Flow

Reachable for SSZ roots and proofs captured after the finalizer sets `pending_checkpoint` at the penultimate block, because `set_pending_checkpoint` does not update the tree and the full rebuild omits the field. The validated start condition is a state/proof artifact whose serialized `pending_checkpoint` differs from the value a consumer expects; live boundary block verification still checks the header `checkpoint_hash` separately.

## Impact

The finalized/proved state root can omit the checkpoint artifact the chain is about to finalize. Restarted, checkpoint-derived, or proof-consuming nodes can accept a state root that does not commit to the pending checkpoint digest used at the epoch boundary, weakening the root-binds-state invariant.

## Root Cause

`pending_checkpoint` is part of serialized consensus state but is excluded from the SSZ state tree leaves and is not reflected by `set_pending_checkpoint`.

## Code

- `ConsensusState` stores `pending_checkpoint`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L30.
- `set_pending_checkpoint` mutates it without updating the SSZ tree: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L268.
[...]

PR #292, opened 2026-06-01
fix: bind pending_checkpoint into the SSZ state root

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/291.

Addresses https://github.com/SeismicSystems/summit/issues/258.

Changes:
- Add `PENDING_CHECKPOINT` as a single scalar leaf: the checkpoint digest when present, the zero hash when absent. No subtree or proof support - the field only needs to be bound into the root, not proven on-chain, and the digest already commits the checkpoint data via SHA-256. The tree takes the digest bytes, staying decoupled from the `Checkpoint` type.
- `set_pending_checkpoint` / `take_pending_checkpoint` now update the leaf, and `rebuild_ssz_tree` passes the digest, so `capture_state_root` reflects it.

### DIFF
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -36,7 +36,7 @@ The state tree is a two-level design: a fixed top-level tree containing scalar f
 
 ### Top-Level Tree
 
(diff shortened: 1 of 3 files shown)


sebastian-osec:
LGTM. This binds the omitted consensus-state field into the rebuilt and captured SSZ state root, so changing that field now changes the advertised root.

Issue #256, opened 2026-05-22
[Security][tentative] Medium (attackable): SSZ State Root Omits Dynamic Epoch Schedule


evonide:
## Context

Summit's consensus state carries a dynamic epoch schedule that maps block heights to epochs after epoch-length protocol changes. Finalizer code uses that schedule to decide epoch-boundary behavior such as last-block handling, pending checkpoint creation, deferred withdrawals, and epoch advancement.

The intended root invariant is that the SSZ state root authenticates every consensus-state field needed by consumers that verify checkpoint state or SSZ proofs. That includes fields that determine which block heights are treated as epoch boundaries.

## Claim

The SSZ state root omits the serialized dynamic epoch schedule even though finalizer boundary logic depends on it, so identical roots can represent different height-to-epoch mappings.

A malicious proof provider or root-only state provider can present a state root that authenticates the ordinary SSZ-tree fields while supplying a different serialized dynamic epoch schedule, causing consumers that trust the root for epoch-boundary state to accept unauthenticated schedule behavior.

## Flow

The path is reachable after epoch-length protocol updates because those updates mutate `DynamicEpocher` without changing an SSZ-tree leaf. The start condition is a consumer that relies on the SSZ state root or SSZ proofs to authenticate schedule-dependent state; this should not be read as proof that Summit's normal checkpoint import accepts arbitrary schedule bytes under the same root, because that path also binds raw checkpoint bytes through the checkpoint digest and finalized-header chain.

## Impact

Root or proof consumers can authenticate a state root without authenticating the epoch schedule that drives checkpoint timing, deferred withdrawal handling, and epoch advancement. A stale or malicious schedule paired with the same root can cause such consumers to classify heights as different epoch-boundary positions, but the normal checkpoint-import path is not claimed to install that schedule solely from the root.

## Root Cause

The dynamic epoch schedule is serialized in `ConsensusState` and used by finalizer logic, but `SszStateTree::rebuild` has no field for it and epoch-length updates do not update any committed SSZ leaf.

## Code

- `ConsensusState` stores `epocher: DynamicEpocher`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L45.
- Epoch-length protocol changes mutate only the dynamic epocher: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L712.
[...]

PR #293, opened 2026-06-01
fix: bind dynamic epoch schedule into the SSZ state root

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/292 (which builds on https://github.com/SeismicSystems/summit/pull/291).

Addresses https://github.com/SeismicSystems/summit/issues/256.

Changes:
- Add `DYNAMIC_EPOCH_SCHEDULE` as a single scalar leaf: the SSZ byte-list root of the encoded epocher. No subtree/proof support, it only needs to be bound into the root, not proven on-chain.
- Set it in `rebuild` (decode/genesis) and refresh it in `capture_state_root`.
  The latter is the key bit: `DynamicEpocher` is `Arc<RwLock<…>>` with `&self` mutators (epoch advance, length update) invoked via `get_epocher()`, bypassing any `&mut ConsensusState` setter, so the commit point is the only reliable place to bind its current value. Both paths use the same hash, so decode-built and live-captured roots agree.

### DIFF
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -36,7 +36,7 @@ The state tree is a two-level design: a fixed top-level tree containing scalar f
 
 ### Top-Level Tree
 
-32 leaf slots (depth 5), 24 used. Each leaf is a 32-byte `hash_tree_root` value. Leaves 24–31 are unused (zero-filled).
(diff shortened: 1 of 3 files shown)


sebastian-osec:
This looks good to me.

Issue #257, opened 2026-05-22
[Security][tentative] Low (attackable): SSZ State Root Omits Keyed Collection Keys


evonide:
## Context

Summit's consensus state stores validator accounts and scheduled validator additions in keyed collections. The keys are consensus-relevant: validator-account keys identify the node public key the account belongs to, and `added_validators` keys identify the epoch when a scheduled activation should be applied.

The SSZ state tree is used to produce state roots and membership proofs for downstream consumers. Its normal invariant should be that a proof binds both a collection value and the key that gives that value consensus meaning.

## Claim

The SSZ tree commits keyed consensus collections as value-only lists, so validator-account and scheduled-activation proofs bind values without binding their node-pubkey or epoch keys.

A malicious SSZ proof or state-root producer can present a state variant where a validator account or scheduled validator addition has a different map key but the committed value sequence is unchanged; because the SSZ tree commits only the keyed collection values, the producer presents the same root for semantically different consensus states and misbinds proofs about validator accounts or activation epochs.

## Flow

1. A proof or state-root producer starts from consensus state containing `validator_accounts` or `added_validators`.
2. The canonical consensus-state codec treats the map keys as part of the serialized state, but the SSZ tree rebuild uses only the ordered values.
3. The producer can change a validator-account key or scheduled-activation epoch while preserving the value-only positional tree content.
4. The resulting SSZ root and proof branch remain valid for the value, but the root does not authenticate the key that gives the value its consensus meaning.

## Impact

External proof consumers cannot authenticate "this account belongs to pubkey X" or "this activation is for epoch E" from the state root alone. This is distinct from pending-execution-request root omission: these collections are included, but their keys are not.

## Root Cause

The SSZ tree models keyed maps as value-only positional lists, while the canonical storage codec treats the keys as consensus state data.

## Code

- `ConsensusState` stores `validator_accounts` as `BTreeMap<[u8; 32], ValidatorAccount>` and `added_validators` as `BTreeMap<u64, Vec<AddedValidator>>`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L28.
- The consensus-state codec writes validator account keys before account values: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L1014.
- The codec writes `added_validators` epoch keys before their vectors: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L1028.
[...]

PR #295, opened 2026-06-01
fix: bind keyed-collection keys into the SSZ state root

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/293 (which builds on https://github.com/SeismicSystems/summit/pull/292 and https://github.com/SeismicSystems/summit/pull/291).

Addresses https://github.com/SeismicSystems/summit/issues/257.

The SSZ tree committed validator_accounts and added_validators as value-only positional lists, so the BTreeMap keys were not bound by the state root. Two states differing only in a map key produced the same root, and validator-account proofs bound a value without binding the pubkey it belongs to.

Changes:
- validator_accounts: add `node_pubkey` as a 9th field; per-validator subtree grows 8 → 16 leaves (depth 3 → 4). Threaded the map key through set_validator_fields and update_/insert_validator_at_slot; rebuild_validators iterates by (key, account); set_account passes the pubkey.
- added_validators: add `epoch` as a 3rd per-item field; subtree grows 2 → 4 leaves (depth 1 → 2). rebuild_added_validators carries the epoch through the flatten.
- Generalized the whole-item proof gindex from a hardcoded subtree depth to `top_gindex << (sd - log_block + 1)` (`log_block = fields_per_item.ilog2()`), so it adapts to the new depths.

A field/whole-item proof's branch now includes the key, so the proof self-binds identity. ValidatorAccount::hash_tree_root no longer equals the per-validator subtree root (the root now includes the node pubkey).

### DIFF
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -74,7 +74,7 @@ Each collection leaf in the top-level tree holds `mix_in_length(subtree.root(), 
 
 #### Validator Accounts
 
-Each validator occupies 8 contiguous leaves (depth-3 per-validator subtree):
+Each validator occupies 16 contiguous leaves (depth-4 per-validator subtree): 9 fields padded to the next power of two. `node_pubkey` is the `BTreeMap` key (the validator's identity) — committing it as a leaf binds the key into the root and into validator proofs. Leaves 9–15 are zero padding.
 
 | Field Index | Field |
 |-------------|-------|
@@ -86,6 +86,8 @@ Each validator occupies 8 contiguous leaves (depth-3 per-validator subtree):
 | 5 | `has_pending_withdrawal` |
 | 6 | `joining_epoch` |
 | 7 | `last_deposit_index` |
+| 8 | `node_pubkey` (map key) |
+| 9–15 | (zero padding) |
 
 Slot assignment is positional: the i-th entry in `BTreeMap` iteration order occupies leaves `[i*8 .. i*8+7]`. The subtree capacity is always a power of 2, growing/shrinking as validators are added/removed.
 
@@ -141,7 +143,7 @@ A `HashMap<pubkey, (epoch_slot, item_slot)>` index enables O(1) proof lookup by 
 
 #### Added Validators
 
-2 leaves per item (node_key + consensus_key), flattened across all epochs.
+4 leaves per item (depth-2 subtree): 3 fields — `node_key` (0), `consensus_key` (1), `epoch` (2) — padded to 4 (leaf 3 is zero). Items are flattened across all epochs, so the `epoch` field commits each item's activation epoch (the `BTreeMap` key), which would otherwise be lost by the flattening.
 
 #### Removed Validators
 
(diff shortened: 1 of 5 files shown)


evonide:
LGTM! This commits the `validator_accounts` pubkey key and the `added_validators` epoch key together with their values so key-only changes now alter the SSZ state root. The widened proof layout and gindex updates also look consistent with the tests.

Small non-blocking follow-up: it would be useful to expose the new key leaves through `ssz_tree_key.rs` and clean up stale proof-layout comments so proof consumers can request and reason about the committed key fields directly.

matthias-wright:
Update(c8e8cab):
- Expose committed key leaves in ssz_tree_key

sebastian-osec:
LGTM. This fixes the issue by committing the keyed-collection keys into the SSZ state root and proof leaves: `validator_accounts` now includes the `node_pubkey` map key, and `added_validators` now includes the activation `epoch` map key. The latest update also exposes those committed key leaves through `ssz_tree_key.rs`, so proof consumers can request them directly.

Non-blocking nits: a few comments/docs still describe the old proof layout.

- [`docs/ssz-merklization.md:L92`](https://github.com/SeismicSystems/summit/blob/c8e8cabd4a60f0e6ee68eafbc4bb2bbf61da59ba/docs/ssz-merklization.md#L92) still says `[i*8 .. i*8+7]`; should be `[i*16 .. i*16+15]`.
- [`types/src/ssz_state_tree.rs:L317-L320`](https://github.com/SeismicSystems/summit/blob/c8e8cabd4a60f0e6ee68eafbc4bb2bbf61da59ba/types/src/ssz_state_tree.rs#L317-L320) still says 8 leaves / depth-3 / `[i*8 .. i*8+7]`; should be 16 leaves / depth-4 / `[i*16 .. i*16+15]`.
- [`types/src/ssz_state_tree.rs:L888-L889`](https://github.com/SeismicSystems/summit/blob/c8e8cabd4a60f0e6ee68eafbc4bb2bbf61da59ba/types/src/ssz_state_tree.rs#L888-L889) still says 2 leaves / depth-1; should be 4 leaves / depth-2 with `node_key`, `consensus_key`, `epoch`, and zero padding.
- [`types/src/ssz_state_tree.rs:L1160-L1164`](https://github.com/SeismicSystems/summit/blob/c8e8cabd4a60f0e6ee68eafbc4bb2bbf61da59ba/types/src/ssz_state_tree.rs#L1160-L1164) still says the validator subtree root is 3 levels above field leaves and mentions `capacity / 8`; should reflect the 16-leaf / depth-4 layout.
- [`types/src/ssz_state_tree.rs:L1546-L1548`](https://github.com/SeismicSystems/summit/blob/c8e8cabd4a60f0e6ee68eafbc4bb2bbf61da59ba/types/src/ssz_state_tree.rs#L1546-L1548) still says the added-validator subtree root is 1 level above field leaves and branch is 1 element shorter; should be 2 levels / 2 elements.

These are comment/doc nits only. The executable constants and proof logic use the new `16` and `4` widths, and the focused tests pass.

Issue #297, opened 2026-06-01
[Security][tentative] High (attackable): Checkpoint Verifier Lacks Weak Subjectivity Freshness


evonide:
## Context

Ethereum-style proof-of-stake clients still require a recent weak-subjectivity checkpoint or equivalent operator trust anchor. A self-contained historical chain can be internally signed by validators that later exited and no longer have economic weight.

## Claim

`verify_checkpoint_chain` verifies that the supplied header chain is contiguous from genesis and that its terminal checkpoint hash matches the checkpoint bytes, but it does not require an independently supplied weak-subjectivity root/epoch or reject stale checkpoint bundles beyond a weak-subjectivity period.

## Flow

This is a recovery/import boundary, not live P2P forgery. The path requires the operator or automation to choose the malicious/stale checkpoint. Summit's withdrawal delay is short, so historical validator keys can become economically detached soon after signing old finalized states.

## Impact

A node can recover onto an attacker-selected long-range history that passes Summit's local cryptographic checks. That can poison startup state, peer authorization, finalized headers, and later proposal or proof state until the operator replaces the checkpoint with a separately trusted recent one.

## Root Cause

Summit's checkpoint verifier treats an internally linked and signed chain from genesis as sufficient trust, but it does not model the weak-subjectivity assumption: the verifier needs a recent trusted root/epoch and a freshness check external to the supplied history.

## Code

- `--checkpoint-path` accepts a local checkpoint path, not a trusted `root:epoch` anchor: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L133-L135
- Startup verifies only when the checkpoint bundle contains a finalized-header chain: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L278-L282
- `verify_checkpoint_chain` starts from genesis validators and walks supplied epoch headers: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/checkpoint.rs#L185-L247
- The verifier checks only terminal checkpoint digest and validator-set consistency after walking the supplied chain: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/checkpoint.rs#L263-L325
[...]

PR #369, opened 2026-06-03
fix: adds weak-subjectivity anchoring for verified checkpoint imports

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/297.

Changes:
  - Requires --weak-subjectivity-epoch and --weak-subjectivity-header-digest when importing checkpoint bundles with finalized_headers/
  - Verifies the supplied finalized-header chain passes through the trusted header digest
  - Rejects checkpoint bundles whose terminal epoch is more than 5 epochs after the trusted anchor
  - Adds getFinalizedHeaderDigest(epoch) RPC for fetching { epoch, digest } from a trusted node
  - Updates checkpointing docs to clarify that self-contained finalized-header bundles are not trustless without an external anchor

### DIFF
--- a/docs/checkpointing.md
+++ b/docs/checkpointing.md
@@ -110,7 +110,7 @@ Checkpoints and headers are stored in a single QMDB store using prefixed keys an
 
 ## Checkpoint Verification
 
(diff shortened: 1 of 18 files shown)


evonide:
LGTM! This fixes the issue by requiring verified checkpoint imports to pass through a recent independently trusted finalized-header digest instead of accepting a self-contained signed history from genesis.

Issue #299, opened 2026-06-01
[Security][tentative] High (correctness): Minimum Stake Change Can Empty Live Committee


evonide:
## Context

The expected invariant is that a valid parameter change should not accidentally transition the live chain into an impossible next-epoch committee. This finding covers a non-empty stake interval where the configured `MinimumStake` is valid in isolation, but no currently active validator satisfies it.

## Claim

An authorized protocol-parameter update raises `MinimumStake` above the balances of all active validators while still leaving a syntactically valid stake range. At the next epoch boundary, Summit applies the new minimum, treats every active validator as below bounds, and removes them from the active validator set used for subsequent consensus operation. The finalizer allows an authorized `MinimumStake` update that is higher than every active validator balance, then independently marks all below-minimum validators inactive and zeroes their balances without checking that at least one active validator remains for the next epoch.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

An accepted minimum-stake change that leaves the active set below the consensus quorum can stop future finalization. Honest nodes may keep running, but the chain cannot make finalized progress until a viable validator set is restored or the unsafe parameter change is repaired.

The practical impact is consensus liveness failure from an authorized parameter transition that Summit accepts without checking the resulting committee size. Recovery requires operator or governance intervention to restore enough active validators or revert the stake-bound change.

## Root Cause

The implementation misses the invariant check or state update described in the claim, allowing the accepted state to be consumed later as if it were valid.

## Code

- Epoch transition applies queued protocol-parameter changes before rebuilding validator lifecycle state: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L544, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L548.
- Applying `MinimumStake` directly updates the state's minimum and reports that stake bounds changed: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L698, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L702.
[...]

PR #370, opened 2026-06-03
fix: prevent stake-bound change from emptying the active validator set

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/279.

Addresses https://github.com/SeismicSystems/summit/issues/299.

Changes:
- Add `prospective_minimum_validator_count()` which returns the floor that will apply next epoch, accounting for a queued `MinimumValidatorCount` change.
- `can_accept_active_validator_exit()` reads the prospective floor instead of the current value, so a same-epoch raise can't be undercut (and a lowering doesn't over-restrict). This flows to both gated paths: stake-bound force-removal staging and voluntary-exit admission.
- Add test `exit_floor_honors_queued_minimum_validator_count_raise` (queued raise tightens, queued lowering loosens, no-change falls back to current).

### DIFF
--- a/types/src/consensus_state.rs
+++ b/types/src/consensus_state.rs
@@ -312,6 +312,26 @@ impl ConsensusState {
         })
     }
 
+    /// Returns the minimum validator count that *will* apply after the queued
+    /// protocol-parameter changes are drained at the next epoch boundary. If no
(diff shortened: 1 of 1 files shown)


evonide:
LGTM please recheck after rebase to make sure the fix survived unchanged.

Issue #300, opened 2026-06-01
[Security][tentative] High (correctness): Seismic Protocol Param Requests Can Make Honest Payloads Fail NewPayload


evonide:
## Context

Summit treats the execution client's `getPayloadV4` response as the canonical source of both the execution payload and the EIP-7685 execution-request list. Validators later submit those same request bytes back to the execution client through `engine_newPayloadV4`.

Seismic adds a type `0xff` protocol-parameter request alongside Ethereum request types. The Engine API still requires request-list elements to be sorted by request type.

## Claim

An authorized or otherwise accepted Seismic protocol-parameter event lands in the same Prague-active block as at least one EIP-7002 withdrawal request or EIP-7251 consolidation request. The proposer receives the malformed ordering from its own execution client, broadcasts the block, and honest validators later call `newPayloadV4` with that raw list. Seismic EVM appends protocol-parameter request type `0xff` before post-execution withdrawal and consolidation system requests. If a block contains both a protocol-parameter request and a withdrawal or consolidation request, Summit stores an honest local payload whose request list is descending by type, and Seismic Reth rejects it as `OutOfOrderExecutionRequest` when peers validate or finalize the block.

## Flow

The path depends on Seismic Reth/REVM integration, not arbitrary byte injection. Source review shows Summit preserves the execution client's request order, Seismic EVM can produce `0xff` before `0x01` or `0x02`, and Seismic Reth validates ascending order before executing the payload.

## Impact

An honest proposer can create a block that Summit consensus accepts structurally but the execution layer rejects during validation or finalization. Affected validators can stop finalizer execution on that canonical mixed-request block instead of cleanly rejecting the invalid payload.

The practical impact is a liveness failure from normal EL/CL integration: the block is produced by the local execution client, accepted by Summit's structural checks, and only fails later when replayed through `newPayloadV4`.

## Root Cause

Seismic's execution-request builder appends the custom high-valued request type before lower-valued post-execution system requests, and Summit does not normalize or pre-validate the list before committing it to a block.

## Code

- Summit stores the `getPayloadV4` execution requests verbatim in the proposed block: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/application/src/actor.rs#L540-L557
[...]

PR #371, opened 2026-06-04
fix: enforce EIP-7685 execution-request ordering in consensus

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/300.

Changes:
- On block propose: sort execution_requests ascending by type byte before computing the block digest, so a proposed block is always spec-compliant with EIP-7685. This is safe because the EL's requests_hash sorts internally, so the committed EL block hash is unaffected (see [here](https://github.com/SeismicSystems/seismic-reth/blob/e8caed73e93c29ad0c116172dc49ed2cfe40f40e/crates/seismic/evm/src/build.rs#L82) and [here](https://github.com/alloy-rs/alloy/blob/55b09e5421a21111103bce4d2b71760a65b939de/crates/eips/src/eip7685.rs#L95)).
- On block verify: reject blocks whose request list is not strictly ascending by type byte.
- Add unit tests.
- Remove the test `test_last_block_withdrawal_before_same_block_deposit_not_dropped` which was testing an edge case where a withdrawal (0x1) lands before a deposit (0x0) in the same block. But this can no longer happen now. This test was added in https://github.com/SeismicSystems/summit/pull/274 to verify the fix for https://github.com/SeismicSystems/summit/issues/248.

Note: since execution request ordering is the responsibility of the EL, I'm proposing this change https://github.com/SeismicSystems/seismic-evm/pull/53 (out of scope for this audit, I'm just linking it for reference).


sebastian-osec:
This looks good to me.

Issue #364, opened 2026-06-01
[Security][tentative] Medium (attackable): Withdrawn Joining Validator Keeps Peer And Observer Authorization


evonide:
## Context

The joining-validator withdrawal path removes the pending activation and queues withdrawal, but leaves the account status as `Joining`.

## Claim

A malicious joining validator can submit a valid full withdrawal before activation; Summit cancels the pending activation but writes the account back as zero-balance `Joining` with a pending withdrawal, so live epoch-transition oracle updates still include the validator's primary key and derived observer keys even though it will never enter the committee. The joining-validator withdrawal path removes the pending activation and queues withdrawal, but leaves the account status as `Joining`.

## Flow

The path requires a validator that has processed a new-validator deposit and is still in its warm-up period, plus control of the matching withdrawal credentials. The stale authorization is strongest on continuously running nodes. A restart drops the stale key through the active-only startup path for the current track, but a later epoch transition can re-add it if the zero-balance `Joining` account still exists.

## Impact

Canceled validators can remain authorized in the live peer and observer set during the pending-withdrawal window. This is stale P2P authorization, not direct voting authorization: Simplex still uses active validators only. The likely impact is bounded by existing P2P/observer admission limits and the withdrawal-completion delay unless validation shows the stale key can fill consensus subchannels or degrade connectivity more directly.

## Root Cause

The cancellation branch updates pending activation and withdrawal bookkeeping but does not change `ValidatorStatus::Joining` to a non-joining status.

## Code

- Joining withdrawal cancellation removes the pending activation when `joining_epoch > current_epoch`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1639-L1653.
- The same branch then zeroes balance, sets `has_pending_withdrawal`, and stores the account without changing status: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1674-L1677.
- Epoch transition updates the network oracle from active-or-joining validators and derived observer keys: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L655-L671.
- Initial finalizer startup tracks active validators only, which explains the temporary restart clearing behavior: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L234-L251.
[...]

PR #373, opened 2026-06-05
 test: cover joining-validator withdrawal exclusion from oracle tracking

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/364.

Changes:
- Adds a regression test for joining-validator withdrawal exclusion from oracle tracking


sebastian-osec:
This regression test looks good to me.

Issue #358, opened 2026-06-01
[Security][tentative] Medium (correctness): Old-Round Syncer Subscriptions Can Panic Application Tasks


evonide:
## Context

Application proposal and verification tasks request blocks from the syncer through subscriptions, then continue once the syncer returns the requested parent or current block. The application treats subscription cancellation as an internal invariant violation rather than a normal missing-data result.

The syncer intentionally drops some old-round subscriptions when the requested missing block may never arrive, such as during notarization or nullification races. That creates a cross-component contract boundary: if the application reaches a proposal or verification path for an already-processed round and the block is absent locally, the syncer can cancel a request that the application currently assumes cannot be canceled.

## Claim

A stale proposal or verification path can cause application work to request a missing block for a round older than `last_processed_round`; the syncer then drops the response path instead of serving, fetching, or registering the subscription. Application propose and verify paths unwrap syncer block subscriptions that the syncer can cancel for missing old-round blocks.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

A stale-round missing-block condition can turn a recoverable subscription cancellation into a panic in application work. Depending on runtime panic handling, this can kill the task, abort proposal/verification for a valid block, or destabilize the node during recovery.

## Root Cause

The syncer represents old-round missing-block subscriptions by dropping the response path, while application callers model cancellation as an impossible invariant and unwrap/expect it.

## Code

- The syncer skips old-round missing-block subscriptions without sending a response: https://github.com/SeismicSystems/summit/blob/ed2c5c8/syncer/src/actor.rs#L681, https://github.com/SeismicSystems/summit/blob/ed2c5c8/syncer/src/actor.rs#L695.
- Verification spawns a task and unwraps the combined parent/block subscription result: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L261, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L264.
- Proposal parent fetch expects the subscription sender to remain alive: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L378.
[...]

PR #375, opened 2026-06-05
fix: remove unwraps/expects on canceled syncer subscriptions

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/358.

Changes:
- Instead of unwrapping/expecting the syncer subscription, return an error and abort the block proposal/verification.


sebastian-osec:
This looks good to me.

Issue #328, opened 2026-06-01
[Security][tentative] Medium (correctness): Forkchoice Update Results Are Ignored


evonide:
## Context

The finalizer is the boundary between Summit consensus state and the execution client. When a finalized or notarized block becomes usable on the consensus side, the execution layer is expected to accept the matching forkchoice before that head is exposed for proposals, verification, or auxiliary data.

The forkchoice response is the execution client's acknowledgement of that handoff. A non-valid response means the consensus head and local execution head are not yet aligned.

## Claim

During honest finalized or notarized block processing, the local execution client can return `SYNCING`, `INVALID`, or another non-valid forkchoice status, but the finalizer still advances consensus readiness and notifies waiters as if the execution layer had adopted the head. The finalizer violates the CL/EL handoff invariant by ignoring the status returned from finalized and notarized forkchoice updates. It can notify consensus waiters and expose a head even when the execution client returned `SYNCING`, `INVALID`, or another non-valid status instead of adopting that head.

## Flow

Reachable when `commit_hash` returns `SYNCING`, `INVALID`, or another non-valid Engine API status after finalized or notarized block processing.

## Impact

After downtime, missing parent headers, or an EL sync lag, the finalizer may treat a block or fork as ready while Reth rejected or deferred the forkchoice. Proposers and verifiers can then request aux data and build on a forkchoice that the EL has not accepted, causing CL/EL divergence, proposal failures, or local stalls.

## Root Cause

The finalizer discards the result of `commit_hash` and continues with consensus notifications instead of gating readiness on an accepted execution-layer forkchoice.

## Code

- Finalized block handling calls `commit_hash` and ignores the result: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L478.
- Notarized fork handling also ignores the `commit_hash` result: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L826.
- The Reth client exposes `ForkchoiceUpdated`, including invalid or syncing states: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L177.
- The finalizer notifies block-height waiters after these paths: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L495, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L846.
[...]

PR #376, opened 2026-06-06
fix: gate execution-layer forkchoice status during block execution

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/328.

Changes:
- Move the forkchoice commit into `execute_block`, right after the `newPayload`-VALID check and before any consensus-state mutation, and classify the status:
`VALID`   -> proceed
`SYNCING` -> `ExecuteOutcome::Syncing` (reuses the existing non-blocking buffer +
             drain retry; safe because nothing is mutated yet)
`INVALID` -> `ExecuteOutcome::InvalidPayload` (finalized: shut down; notarized:
             discard the fork)
transport error -> fatal (unchanged)

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -912,6 +912,8 @@ impl<
                 &mut self.canonical_state,
                 &self.protocol_consts,
                 self.deposit_signature_domain,
+                // canonical path: instant finality (safe = finalized = head)
+                None,
             )
             .await
             {
@@ -1000,17 +1002,8 @@ impl<
             );
         }
 
-        if let Err(e) = self
-            .engine_client
-            .commit_hash(*self.canonical_state.get_forkchoice())
-            .await
-        {
-            error!(target: "critical", "engine client error on canonical commit_hash after finalization: {e}");
-            #[cfg(feature = "prom")]
-            counter!("critical_errors_total", "reason" => "engine_client_error", "severity" => "critical")
-                .increment(1);
-            return Err(anyhow!("engine client error on canonical commit_hash: {e}"));
-        }
+        // Forkchoice was already committed (and its status gated) inside
+        // `execute_block` before any state mutation.
 
         #[cfg(feature = "prom")]
         {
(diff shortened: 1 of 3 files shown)


sebastian-osec:
This fixes the direct finalized execution path, but I think the notarized-then-finalized path still skips the finalized forkchoice update.

For a notarized block, `execute_block` calls `commit_hash` with `head = B` but `safe = old_canonical_finalized` and `finalized = old_canonical_finalized`, which is correct because the EL cannot treat a speculative fork as finalized. If that returns `VALID`, the block is stored in `fork_states`.

Later, when the same block becomes finalized, `handle_finalized_block` sees it already exists in `fork_states` and reuses that state instead of calling `execute_block` again. Since this PR removed the later `commit_hash`, the finalizer never sends/gates the canonical finalized forkchoice update with `head = B, safe = B, finalized = B` before notifying waiters/acking finalization.

So the PR gates speculative fork adoption, but not promotion of that fork to finalized/canonical.

matthias-wright:
Update(c341f36):
- Fork state reuse branch in handle_finalized_block now sends + gates the finalized forkchoice before promoting the fork state to canonical:
    - Syncing    → buffer & retry (canonical state untouched, replays cleanly)
    - InvalidPayload → fatal (EL/CL inconsistency on a finalized block)
    - engine Err → fatal

sebastian-osec:
The main fix looks right to me for finalized and notarized block execution.

`commit_hash` is now gated before the finalizer exposes the block as ready, and the notarized-then-finalized reuse path now sends the finalized forkchoice before promoting the fork state. That addresses the earlier gap where a fork could be treated as canonical without the EL accepting `head = safe = finalized = B`.

One thing I’d like to clarify before treating #328 as fully closed: startup from a checkpoint when the initial `commit_hash` returns `SYNCING`.

This PR lets the finalizer enter the main loop in that case, which is good for mailbox responsiveness. But it also means the finalizer can still serve aux data from the checkpoint-loaded canonical state before the EL has accepted that checkpoint forkchoice.

If that’s intended, meaning startup responsiveness is preferred and later block execution is the place where EL catch-up gets gated, then this looks fine to me. If #328 means “don’t expose aux data for any head until the EL has accepted the matching forkchoice,” then I think this startup case is still a small remaining gap.

matthias-wright:
Good point. I think serving aux data while the EL returns SYNCING is an acceptable trade-off for maintaining mailbox responsiveness. If the served aux data causes an invalid block, it will be rejected by the peers.

Issue #374, opened 2026-06-05
[Security][tentative] Critical (correctness): Last-block top-up after stake-bound staging can drop validator balance


sebastian-osec:
## Context

Summit stages stake-bound validator removals at the penultimate block of an epoch so the last block's finalized header contains the validator-set delta. The actual protocol-parameter update and stake-bound withdrawal handling happen at the epoch boundary.

Validator top-ups are accepted as execution requests and may be queued for later deposit processing. Existing-validator top-ups should preserve the current validator balance and add the top-up amount.

## Claim

A valid `MinimumStake` increase can stage an active validator for removal at the penultimate block, then a last-block top-up from that same validator can prevent Summit from scheduling the stake-bound withdrawal for the validator's existing balance.

After the epoch boundary, the staged removal marks the account `Inactive`. When the queued top-up is processed later, Summit treats the account as a new-validator deposit and computes the balance from only the top-up amount, ignoring the pre-existing stake.

## Flow

1. A valid protocol-parameter update raises `MinimumStake` above an active validator's current balance.
2. At the penultimate block, Summit sees the validator will be below the prospective minimum and pushes it into `removed_validators`.
3. At the last block, the validator submits a valid top-up deposit.
4. Deposit parsing marks the account `has_pending_deposit = true`.
5. The epoch-boundary committee update applies `removed_validators` and marks the account `Inactive`.
6. The stake-bound withdrawal scan skips the account because `has_pending_deposit` is true.
7. At the next deposit-processing point, the pending top-up is processed against an `Inactive` account, so Summit uses `new_balance = request.amount` instead of `account.balance + request.amount`.
8. If the top-up alone is below the new minimum, Summit refunds only the top-up and removes the account. The original balance is not retained and no withdrawal is queued for it.

## Impact

This can drop a validator's existing bonded balance from Summit consensus accounting without scheduling a corresponding EL withdrawal.

In the reproduced scenario, a validator with `32 ETH` was affected by a valid `MinimumStake(33 ETH)` update and submitted a `1 ETH` last-block top-up. Summit later refunded only the `1 ETH` top-up and removed the account, with no withdrawal for the original `32 ETH`.

This is a correctness issue under normal authorized operation: it requires a valid stake-bound protocol-parameter update and a validator-controlled top-up, not arbitrary request forgery.

## Regression Note

This appears to come from the interaction of two otherwise reasonable fixes:

[...]

PR #377, opened 2026-06-08
fix: conserve staged-removal validator balance on deferred top-up

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/276 (which builds on https://github.com/SeismicSystems/summit/pull/275).

Addresses https://github.com/SeismicSystems/summit/issues/374.

Changes:

- Guard the Inactive deposit branch on `account.balance > 0` to distinguish a staged-removal validator (still holds a credited balance) from a new-validator placeholder (balance 0). For the former, withdraw the full credited balance (`balance_deduction = balance`), refund the top-up separately, and let the completion path remove the account. Placeholders and active top-ups are unaffected.
- Add regression test `test_last_block_topup_does_not_drop_staged_removal_balance`.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -3018,17 +3018,26 @@ async fn process_execution_requests<
                             bonded_balance,
                         );
 
-                        // Refund the top-up, which was never credited to the balance
-                        // (taxed like any other invalid-deposit refund).
+                        // Refund the top-up in full, untaxed. Unlike the
+                        // invalid-deposit refund paths, this top-up was valid in
+                        // shape, signature, and resulting balance (it passed
+                        // admission and would have landed in range) — it is refunded
+                        // only because the independent stake-bound removal wins. The
+                        // depositor is blameless, the bonded balance above is also
+                        // returned untaxed, and the canonical stake-bound exit applies
+                        // no tax, so `invalid_deposit_tax` must not apply here. The
+                        // top-up was never credited to the balance, so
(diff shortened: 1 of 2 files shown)


sebastian-osec:
Question: should the top-up refund in this staged-removal path be taxed as an invalid deposit? The top-up passed validation when accepted; it is refunded only because the earlier staged removal wins. Since this branch uses `push_invalid_deposit_withdrawals`, a nonzero `invalid_deposit_tax` sends part of the valid top-up to treasury. If that is intended policy, this is fine, but the regression test only covers tax = 0.

matthias-wright:
Good question. I think a tax-free withdrawal is more appropriate here.

Update(ca07b67):
- Refund staged-removal top-up untaxed

sebastian-osec:
Thanks. If we are okay with `max_withdrawals_per_epoch` applied once to validator withdrawals and once to refund withdrawals, then my concern is resolved. #276 .

One small follow-up: the docs still describe `max_withdrawals_per_epoch` as limiting total withdrawals in the last block of each epoch, not as a per-type limit:
https://github.com/SeismicSystems/summit/blob/ca07b67fb08a93218221260db48a65832549a414/docs/deposits-and-withdrawals.md#L123

The latest commit addresses my original tax concern too. The staged-removal top-up refund now bypasses the invalid-deposit tax helper and refunds the full top-up amount, with a nonzero-tax regression test covering it.

Issue #352, opened 2026-06-01
[Security][tentative] Medium (attackable): Speculative Fork States Can Mark EL Forks Finalized


evonide:
## Context

The finalizer tracks consensus-layer fork states and translates the selected state into Engine API forkchoice data for execution payload building. A notarized fork may be safe to execute speculatively before it is finalized, but the execution layer's finalized hash must continue to reflect canonical finalized history rather than the speculative fork head.

## Claim

A validator selected to build on a notarized but unfinalized fork can make its local application request an execution payload with aux data copied from the stored fork state, causing the Engine API forkchoice to mark that speculative fork head as the finalized EL block. Stored fork states can expose a notarized but unfinalized fork head as the execution-layer finalized hash in later proposal aux data.

## Flow

The path is reachable when the finalizer executes and stores a notarized fork state, then later the local consensus/application path requests proposal aux data for that fork before it is finalized. A single Byzantine proposer is not enough by itself: the fork state must already have a valid notarization/quorum certificate, and the later proposal parent digest must be the stored fork head rather than the canonical head.

## Impact

A validator building on a merely notarized fork can tell its execution client that the speculative fork head is finalized. If that fork loses, the local EL may be unable to follow the canonical CL-finalized branch or may reject later forkchoice updates that conflict with the incorrectly finalized hash.

## Root Cause

Consensus-state normalization overwrites the stored EL safe/finalized hashes with the speculative state head, and proposal aux data later reuses those stored forkchoice fields for Engine API payload building.

## Code

- Notarized, not-yet-finalized blocks are executed into cloned fork states: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L804.
- The mutated fork state is stored for later use: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L816.
- `execute_block` normalizes every executed state's safe/finalized forkchoice hashes to the state head: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1451.
- The immediate speculative Engine API commit preserves canonical finalized hash instead of finalizing the fork: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L824.
- Proposal aux data later exposes the selected parent fork state's stored forkchoice: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L973.
[...]

PR #378, opened 2026-06-08
fix: don't expose unfinalized fork head as EL-finalized in proposal aux data

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/352.

Changes
- `handle_aux_data_mailbox` now exposes `{ head: parent head, safe/finalized: canonical finalized }` instead of the stored fork forkchoice. No-op for canonical parents; stops advertising a speculative head as finalized for fork parents.
- Add regression test `test_fork_aux_data_does_not_finalize_unfinalized_fork_head`.


sebastian-osec:
This looks good to me.

Issue #339, opened 2026-06-01
[Security][tentative] Medium (correctness): Orphaned Queued Top-Up Can Rebind To Replacement Validator Account


evonide:
## Context

Stake-bound account removal can delete the account context for a queued top-up without deleting the queued deposit, letting that stale top-up bind to a later replacement account for the same node pubkey.

## Claim

Stake-bound account removal can delete the account context for a queued top-up without deleting the queued deposit, letting that stale top-up bind to a later replacement account for the same node pubkey.

## Flow

The path requires a top-up to stay in `deposit_queue` long enough for a min/max stake change, stake-bound removal, withdrawal completion, and account deletion, likely through deposit backlog or a low/zero deposit-processing cap. The replacement deposit must arrive after the old full-withdrawal block removes the account; if it arrives before deletion, the existing account's `has_pending_withdrawal` flag causes deposit rejection/refund instead of replacement account creation.

## Impact

The old top-up can be interpreted under the wrong account lifecycle. It can be popped against a replacement account with different withdrawal credentials and consensus key, enter refund/removal handling using replacement metadata, or be skipped if processed after deletion but before replacement.

## Root Cause

The implementation misses the invariant check or state update described in the claim, allowing the accepted state to be consumed later as if it were valid.

## Code

- Accepted deposits are queued separately after marking or creating the account: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1492-L1537.
- The deposit queue stores and pops raw requests independently of account lifetime: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L509-L537.
- Stake-bound enforcement can mark an account inactive, zero balance, and queue withdrawal: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1218-L1264.
- Zero-balance accounts are removed when the scheduled withdrawal is processed: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1888-L1913.
- `remove_account` removes the account but does not purge same-pubkey queued deposits: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L430-L444.
[...]

PR #379, opened 2026-06-08
chore: add consensus-key guard when processing deposits

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/377 (which builds on https://github.com/SeismicSystems/summit/pull/276 and https://github.com/SeismicSystems/summit/pull/275).

Addresses https://github.com/SeismicSystems/summit/issues/339.

Changes:
- Adds consensus-key guard when processing deposits.

Note: This guard is purely defensive,
1. Parse already enforces the match. A top-up is only accepted into the queue if its consensus key equals the stored account's. For a new-validator deposit, the placeholder's consensus key is set from the deposit, so they match trivially.
2. The account can't change identity before consume. While the deposit is queued, has_pending_deposit = true, which blocks the account from being removed/replaced. So the account that's looked up at consume time is the same account, with the same consensus key, that was validated at parse.


matthias-wright:
Continued here https://github.com/SeismicSystems/summit/pull/380

Issue #321, opened 2026-06-01
[Security][tentative] Medium (correctness): Engine Transport Reconnect Can Deadlock Finalizer


evonide:
## Context

The finalizer executes and commits payloads through the Reth Engine API while also servicing consensus, syncer, RPC, and cancellation messages through its actor loop.

## Claim

Reth IPC is unavailable, wedged, or repeatedly disconnected while finalizer processes a finalized or notarized block that requires Engine API interaction. Engine transport errors enter an indefinite reconnect loop without timeout or cancellation, and finalizer calls those methods inline while processing blocks.

## Flow

This is currently a local dependency failure rather than a proven peer-triggered attack. Escalation would require showing a malicious block or peer can reliably crash or wedge Reth transport.

## Impact

Finalizer can stop servicing its mailbox, preventing block processing, aux-data responses, acks, epoch transitions, and cancellation handling.

## Root Cause

Transport reconnect is an uncancelable infinite retry inside actor-critical Engine API calls.

## Code

- Transport reconnect loops until available: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L73.
- Engine methods enter reconnect on transport errors: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L115, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/engine_client.rs#L161.
- Finalizer calls execution and commit inline: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L438, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L478, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L805, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L831.
- Finalizer cancellation is only checked in the outer actor loop, so an awaited Engine API call can prevent returning to mailbox and cancellation handling: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L306.
- Finalizer mailbox queries and syncer updates share the same actor progress: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/ingress.rs#L17, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/ingress.rs#L481.

## Related Issues/PRs

[...]

PR #381, opened 2026-06-08
chore: bound the number of IPC reconnect attempts in the engine client

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/321.

Changes:
- Bounds the number of IPC reconnect attempts in the engine client to 200. The engine client waits 100 millis between each attempt, so the total reconnect interval is around 20 seconds.

Note: The finalizer is still deadlocked during these reconnect attempts, but that is acceptable. If the finalizer cannot connect to the engine API, it cannot perform meaningful work. Serving RPC requests is secondary in priority.

### DIFF
--- a/types/src/engine_client.rs
+++ b/types/src/engine_client.rs
@@ -31,6 +31,10 @@ use alloy_transport::TransportError;
 use alloy_transport_ipc::IpcConnect;
 use std::future::Future;
 
+/// The number of times the engine client will try to reconnect
+/// after failing to connect to the IPC socket.
+const IPC_CONNECT_MAX_RETRIES: u32 = 200;
+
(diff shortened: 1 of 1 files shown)


sebastian-osec:
Nit: this now bounds the reconnect loop, which addresses the infinite-wait part of #321. It may be clearer for `wait_until_reconnect_available()` to return a `Result` and explicitly report “reconnect attempts exhausted” instead of returning `()` after both success and exhaustion. Today callers retry against `self.provider` after the helper returns, so exhaustion is only observed indirectly through that second Engine call failing.

matthias-wright:
Update(4c2a7a4):
- Report reconnect exhaustion explicitly in engine client

sebastian-osec:
Looks good to me.

Issue #326, opened 2026-06-01
[Security][tentative] Medium (correctness): Finalizer DB Errors Advance Epoch Side Effects


evonide:
## Context

The finalizer persists finalized headers, checkpoints, consensus state, and commit markers at epoch boundaries. Those writes are the recovery boundary for the node's canonical state, while the same handler also updates in-memory epoch state, peer-oracle membership, and orchestrator epoch engines.

Database errors are treated as fatal by cancelling the shared cancellation token. The intended invariant is that once a fatal persistence error is observed, the current consensus-critical transition should stop or return an error before later components are told to run on the uncommitted state.

## Claim

During normal epoch-boundary finalization, a storage failure in `store_finalized_header`, `store_finalized_checkpoint`, `store_consensus_state`, or `commit` can cancel the node but remain invisible to the in-flight `handle_finalized_block` call. Until the actor loop observes cancellation after the handler returns, the same node can advance local epoch side effects and start or stop epoch engines even though the durable finalizer transition failed. Finalizer database write and commit errors only log and cancel shutdown; their APIs return `()` and the epoch-boundary handler continues into in-memory epoch mutation, oracle tracking, orchestrator `Enter`, and old-epoch `Exit`.

## Flow

The path is reachable if the local QMDB/storage layer returns an error during an epoch-boundary write or commit. The trigger is a local storage fault, filesystem failure, or injected backend error, not a remote peer directly controlling database results.

## Impact

An affected node can briefly expose or act on next-epoch peer-oracle and orchestrator state that was not durably committed. On restart, durable finalizer state may lag the already-reported local epoch side effects, increasing recovery inconsistency and making operators rely on a critical log/cancellation rather than a fenced transition.

## Root Cause

The finalizer DB wrapper converts fatal persistence errors into cancellation side effects instead of returning `Result` to the epoch-boundary state machine.

## Code

- `handle_db_error` logs a critical database error and cancels the token, but returns no error to callers: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/db.rs#L50-L55
[...]

PR #382, opened 2026-06-09
fix: fence epoch-boundary side effects on fatal persistence failure

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/326.

Changes
- `handle_db_error` now returns the error; the four boundary write/commit methods (and the internal `set_latest_*_epoch` tracker writes) return `anyhow::Result<()>` and propagate it. Tests updated.
- In the finalizer actor, `fatal persistence errors return `Err` before any oracle/orchestrator side effect and the existing `Err => cancel(); break` arm shuts down cleanly.


sebastian-osec:
This looks good to me as well.

Issue #323, opened 2026-06-01
[Security][tentative] Medium (correctness): Finalized Block Delivery Is Not Idempotent


evonide:
## Context

The syncer delivers finalized blocks to the finalizer/application as the canonical ordered stream. Its API explicitly documents at-least-once delivery, so the downstream reporter must tolerate duplicate finalized deliveries without repeating consensus-state, Engine API, or epoch-boundary side effects.

## Claim

A restart, recovery, resolver, or syncer replay path delivers the same finalized block more than once while the block has already been applied to the finalizer's canonical state. The syncer API allows at-least-once delivery, but the finalizer handles the replay as fresh canonical work. `handle_finalized_block` lacks the stale-height/idempotence guard that the notarized path has. If a duplicate `Update::FinalizedBlock` reaches the finalizer, it either installs an old fork state or executes the block again on canonical state, then commits forkchoice and acknowledges the replayed height.

## Flow

The source-level idempotence gap is clear once duplicate finalized delivery occurs. The reviewed syncer source drops newly stored finalizations at or below `last_processed_height` and uses FIFO pending acknowledgements, so ordinary post-ack re-finalization is not the demonstrated trigger. The remaining validation question is which production retry, restart, repair, or acknowledgement/progress-sync failure can re-deliver a finalized block after the finalizer has applied the height but before all relevant syncer/application progress is durably advanced.

## Impact

A duplicate within the same epoch can re-run Engine API payload checking, execution-request parsing, deposit/withdrawal processing, forkchoice commit, and height notifications. A duplicate epoch-boundary block can also re-enter finalized-header, checkpoint, validator-committee, and orchestrator transition work, or hit epoch assertions after canonical state has advanced.

## Root Cause

Finalized delivery is treated as exactly-once at the finalizer boundary even though the syncer contract says finalized blocks are delivered at least once.

## Code

- Syncer documents finalized block delivery as at-least-once and says the reporter should handle duplicates: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/syncer/src/lib.rs#L21-L24.
[...]

PR #383, opened 2026-06-09
fix: make finalized block delivery idempotent in the finalizer

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/323.

Changes:
- Added the idempotence guard at the start of handle_finalized_block (before the fork-state/execute branch).
- Added regression test `test_duplicate_finalized_delivery_is_idempotent`.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -685,6 +685,51 @@ impl<
         let height = block.height();
         let block_digest = block.digest();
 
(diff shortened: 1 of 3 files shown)


sebastian-osec:
This looks good to me. PR #383 adds the finalized-delivery idempotence guard and ACKs duplicates without re-executing the block, so the duplicate finalized-block replay path from #323 is resolved.

Issue #319, opened 2026-06-01
[Security][tentative] Medium (correctness): Engine Actor Failures Are Swallowed


evonide:
## Context

The node engine starts the application, broadcast buffer, finalizer, syncer, and orchestrator actors. Those actors are the consensus-critical runtime; a tracked actor failure should either bring down the node or propagate an error to the outer supervisor so the process can be restarted.

## Claim

A reachable application, finalizer, syncer, buffer, or orchestrator actor failure causes its handle to resolve with `Err`. Instead of propagating that failure to the top-level `try_join_all`, the engine task completes normally while other top-level services may remain alive. `Engine::run` joins the tracked actor handles with `try_join_all`, but if any actor returns `Err`, it only logs `"engine failed"` and then returns `()`. The outer node task sees the engine handle as successful and continues waiting on P2P and RPC handles, so a consensus actor failure can be converted into a clean engine return.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

Operators can be left with a stale node process that still exposes network or RPC services while its consensus engine has stopped. External supervisors that rely on process exit or top-level task errors may not restart the node promptly.

## Root Cause

The engine supervision boundary logs tracked actor errors but does not propagate them, cancel sibling services, or stop the runtime context.

## Code

- `Engine::run` joins application, buffer, finalizer, syncer, and orchestrator actor handles: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/engine.rs#L453-L475.
- In the actor-future branch, `Err` is logged but not returned or rethrown: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/engine.rs#L479-L486.
- Top-level startup only logs failures if `try_join_all(vec![p2p, engine, rpc_handle])` returns `Err`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L425-L428.
- The same top-level join shape is used in the benchmarking startup path: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L620-L623.

## Related Issues/PRs

[...]

PR #384, opened 2026-06-09
fix: bring the node down when any core task exits

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/319.

Changes:
- Added `supervise_node_tasks(p2p, engine, rpc)` which fuses the three task handles and `futures::select!`s so the first to exit logs and returns, letting the runtime tear the process down for restart.
- It is called from both `run_node_inner` and `run_node_local_inner`, replacing the `try_join_all(...)` joins.


sebastian-osec:
This looks good to me.

Issue #335, opened 2026-06-01
[Security][tentative] Medium (attackable): Observer Child Keys Are Not Namespace Bound


evonide:
## Context

Summit supports observer identities derived from validator node keys. The finalizer derives authorized observer public keys from each validator node key and `observers_per_validator`, while node startup uses a child private key to join the authenticated P2P network as an observer.

The network namespace is intended to prevent replay across deployments, but that namespace is used when signing P2P messages, not when deriving the observer child identity that the live peer oracle authorizes.

## Claim

A validator reuses the same Ed25519 node key on two Summit deployments and delegates or leaks observer child key `i` on deployment A. If deployment B authorizes at least `i + 1` observers for that validator node key, the holder of the deployment-A child key can sign deployment B's namespace and be accepted as deployment B's observer identity. Observer child keys are derived from only the validator node key and observer index. The derivation omits the Summit namespace or chain identity, so the same validator node key and observer index produce the same observer private/public key across deployments.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

An observer key issued for one deployment can become an unauthorized observer on another deployment. The direct impact is cross-deployment peer admission and capacity consumption; existing observer subchannel findings show that accepted observer identities can matter for liveness, but this report does not prove BLS consensus signing, funds loss, or finalized-state divergence.

## Root Cause

Observer child key derivation lacks domain separation for the Summit namespace, genesis, or chain identity.

## Code

- Observer child private key derivation uses path `m/seismic/observer/{index}`, the master seed, and master public key, but no namespace: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/ext_private_key.rs#L31-L40.
- Public observer derivation also uses only the master public key and index: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/ext_private_key.rs#L109-L118.
[...]

PR #385, opened 2026-06-10
fix: namespace-separate observer child-key derivation

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/335.

Changes:
- Mix the genesis namespace into the child key derivation.
- Add unit tests.


sebastian-osec:
This fix looks good to me.

Issue #354, opened 2026-06-01
[Security][tentative] Medium (correctness): Stale Epoch Enter Can Use The Current Epoch Genesis


evonide:
## Context

The orchestrator starts per-epoch consensus engines, and each engine asks the finalizer for the genesis digest that roots that epoch's consensus state. During recovery or catch-up, a nonzero epoch-entry message queued by the finalizer can be drained after the finalizer has already advanced again, so genesis lookup must be keyed to the requested epoch rather than only the current canonical state.

## Claim

A recovering or catching-up node can drain an older queued nonzero `Enter(epoch)` after the finalizer has already advanced to a newer canonical epoch; the stale epoch engine asks for its genesis, the finalizer returns the current epoch genesis instead of the requested one, and the node starts old-epoch consensus work rooted in the wrong digest. The finalizer answers epoch-genesis requests from current canonical state instead of serving genesis hashes keyed by the requested epoch.

## Flow

Reachability is reasoned from source, with a narrower start condition than an arbitrary stale `Enter`. The path requires the finalizer to queue an `Enter` for a nonzero recovered/current epoch, then advance to a later epoch before the orchestrator starts that queued epoch, so the finalizer's canonical epoch no longer matches the requested epoch. Epoch 0 is not affected by the same genesis-query path because the application returns the configured genesis hash locally for `epoch == 0`.

## Impact

A recovering or catching-up node can start an old-epoch Simplex engine with the next epoch's genesis digest. That engine can emit old-epoch consensus traffic rooted in the wrong epoch genesis, poison local journal state for that epoch, and stall recovery or epoch handoff until restart or manual cleanup.

## Root Cause

The finalizer stores only the current canonical epoch genesis hash and answers `GetEpochGenesisHash` with that value even when the request names a different epoch.

## Code

- Engine startup starts the finalizer and syncer before the orchestrator drains epoch-entry messages: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L453, https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L455, https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L462.
[...]

PR #386, opened 2026-06-10
fix: serve epoch genesis hash by requested epoch

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/354.

Changes:
- GetEpochGenesisHash now serves by requested epoch:
 equal to current -> canonical genesis; == 0 -> configured genesis hash;
  less than current -> digest of the last block of epoch-1 (from the durable finalized header);
  higher than current (unreached) or missing -> decline (drop the response so the caller errors) instead of returning a wrong genesis.
- Added unit tests.


sebastian-osec:
This change looks good to me.

Issue #359, opened 2026-06-01
[Security][tentative] Medium (correctness): Temporary Verify Dependencies Return Terminal False


evonide:
## Context

The application implements Commonware's block verification callback for consensus proposals. Verification depends on local finalizer state and checkpoint auxiliary data, both of which can lag temporarily even when the proposed block itself is valid.

Commonware treats a `false` verification result as the terminal verdict for that proposal, so the application must keep verification pending while local dependencies can still catch up. Returning `false` for temporary dependency misses converts local timing into a permanent vote rejection.

## Claim

A valid proposer can have its proposal rejected by honest validators whose local finalizer or aux-data dependency is temporarily behind or unavailable during verification. Application verification returns terminal `false` for temporary finalizer lag and missing auxiliary-data dependencies.

## Flow

The path is reachable through `Automaton::verify`. The prerequisite is a timing window where the proposal may be valid but the validator has not executed the parent or cannot yet fetch required aux data.

## Impact

Honest validators can permanently reject an otherwise valid proposal because their local finalizer or aux-data path was late. Under load, restart, or epoch-boundary timing, this can turn local lag into missed notarizations and liveness stalls.

## Root Cause

Verification conflates temporary dependency unavailability with permanent block invalidity by returning terminal `false` for lagging finalizer and missing aux-data conditions.

## Code

- Verification returns false when the parent has not been executed by the finalizer: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L271.
- Verification returns false when aux data is unavailable: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L291.
- The final false response is sent around aux-data failure handling: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L315.

## Related Issues/PRs

The following existing items touch the same trust boundary, adjacent root cause, or likely remediation stack.

- #304
- #358
- #177
- #167

Note that #359 is retryable local dependency lag reported as terminal verification failure.

## Fix

- Distinguish permanent invalidity from temporary unavailability.
- Keep verification pending, retry internally, or surface a retryable/pending response where Commonware supports it.
- Avoid returning false for finalizer lag, missing checkpoint aux data, or local catch-up delay.
- Add tests where aux data becomes available shortly after the first verification attempt.

matthias-wright:
[...]

PR #387, opened 2026-06-10
chore: clarify verify/certify parent-check naming and logs

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/359.

Changes:
- Updates comments and logs related to height notifications for parent blocks in `Message::Certify` and `Message::Verify`.


sebastian-osec:
Yes, I agree with this. After re-checking the notify_at_height path, it does not return false just because the finalizer is temporarily behind. In that case it stores the waiter and resolves it once the matching parent is executed.

So I agree this was a misleading comments/logs issue rather than a behavioral bug in this path.

Issue #227, opened 2026-05-22
[Security][tentative] Medium (attackable): Live Consensus Domain Is Not Chain Bound


evonide:
## Context

Summit uses domains to scope live peer authentication and BLS consensus signatures. Validators run nodes with Ed25519 networking keys and BLS consensus keys; separate deployments may accidentally or intentionally reuse similar configuration.

The intended trust boundary is the local chain deployment: a live peer identity or consensus certificate from one deployment should not verify as evidence for another unless both deployments intentionally share the same chain identity. Domain separation is the mechanism that should bind signatures to that identity.

## Claim

Summit uses the configurable genesis `namespace` directly as both the live P2P authentication namespace and the BLS multisig consensus domain. The namespace is not derived from the EL genesis hash, Summit genesis digest, protocol version, or immutable chain configuration, and the signed block/header digest does not add that chain identity back in.

A malicious deployment operator or peer on another Summit deployment that reuses the same validator node keys, BLS keys, and namespace can make live peer authentication and consensus certificates verify across deployments because the live domains are not derived from immutable chain identity.

## Flow

The path requires two deployments sharing the same namespace and validator node/BLS keys; overlapping or replayable block/header digests increase the consensus-certificate impact.

## Impact

Two deployments that reuse validator node/BLS keys and namespace can authenticate peers and verify live consensus certificates across networks. If histories or payload digests overlap, a foreign notarization/finalization certificate can be accepted as local consensus evidence. Even when histories diverge, this enables cross-network resolver traffic, DoS, and partition risk between networks that operators believe are distinct.

## Root Cause

The live P2P and BLS scheme domains are set directly from the configurable genesis namespace, while signed header digests omit immutable chain identity.

## Code

- Runtime network setup takes `genesis.namespace.as_bytes()` directly: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L372.
- Engine startup builds the consensus scheme provider from the same namespace bytes: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L137.
- `SummitSchemeProvider` creates signer/verifier schemes with only `self.namespace`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/scheme.rs#L124, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/scheme.rs#L141.
[...]

PR #388, opened 2026-06-10
fix: bind live P2P and consensus domains to chain identity

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/227.

Changes:
- Add `chain_domain(genesis_hash, namespace)` in `types/src/lib.rs`: `H("summit-chain-v1" || PROTOCOL_VERSION || genesis_hash || len(ns) || ns)`, mirroring `deposit_signature_domain`.
- Add `Genesis::genesis_hash()` to centralize `eth_genesis_hash` parsing; reuse it in `node/src/config.rs`.
- `node/src/engine.rs`: derive the consensus domain once in `Engine::new` and use it for the BLS scheme provider (signer and verifier) and the syncer/orchestrator namespace fields.
- `node/src/args.rs`: derive the same domain for the P2P authentication namespace at both node startup paths.
- `types/src/checkpoint.rs`: reconstruct the same domain in `verify_checkpoint_chain` so finalization certificates still verify.
- Keep the finalizer and RPC namespaces raw, since `deposit_signature_domain` already folds in the genesis hash separately.
- Add unit tests.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -392,9 +392,12 @@ pub struct Finalizer<
     oracle: O,
     node_public_key: PublicKey,
 
-    /// Chain namespace (genesis), mixed into observer child-key derivation for
-    /// domain separation across deployments.
-    namespace: Vec<u8>,
+    /// Chain bound domain (`chain_domain(config_digest)`) used to derive the
+    /// authorized observer child keys. Must match the domain the live P2P
+    /// observer signer is derived with (see node startup), or observers will
+    /// not be authenticated. Distinct from the raw deposit namespace, which is
+    /// already folded into `deposit_signature_domain`.
+    observer_domain: Vec<u8>,
     validator_exit: bool,
     cancellation_token: CancellationToken,
     _signer_marker: PhantomData<S>,
@@ -511,7 +514,7 @@ impl<
                     &cfg.namespace,
                 ),
                 node_public_key: cfg.node_public_key,
-                namespace: cfg.namespace,
+                observer_domain: cfg.observer_domain,
                 validator_exit: false,
                 cancellation_token: cfg.cancellation_token,
                 _signer_marker: PhantomData,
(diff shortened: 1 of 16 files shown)


sebastian-osec:
The implementation looks consistent to me for the intended `eth_genesis_hash + namespace` scope. P2P auth, consensus signatures, sync/orchestrator verification, and checkpoint verification all use the derived `chain_domain(...)`, while deposit signatures still keep their separate genesis-bound domain.

One question before merge: are we defining chain identity here as only `eth_genesis_hash + namespace + protocol_version`? Issue #227 mentions binding to immutable chain identity / config digest. If two Summit deployments share the same EL genesis hash and namespace but differ in other Summit genesis config, for example validator set / consensus keys, epoch length, timeout config, stake limits, or deposit/withdrawal limits, they would still share the same live P2P/consensus domain. If that scope is intentional, this looks good to me.

matthias-wright:
Update(e12ccd3):
- Bind live P2P/consensus domain to the full genesis config digest

sebastian-osec:
Thanks, the new commit addresses my previous concern. The live P2P/consensus domain is now derived from `Genesis::config_digest()` rather than only `eth_genesis_hash + namespace`, so deployments that differ in validator keys, consensus params, stake limits, epoch config, etc. should no longer share the same live domain.

The main #227 invariant looks covered to me: P2P auth, consensus signing/verifying, syncer/orchestrator verification, and checkpoint verification all reconstruct the same `chain_domain(genesis.config_digest())`. The added tests also cover per-field digest separation and cross-domain signature replay rejection.

Small nit: observer metadata/test code still derives the observer child public key with raw `genesis.namespace`, while actual observer P2P startup now derives the signer using the chain-bound domain. That does not look like a #227 bypass, but it may report/check the wrong observer public key and should probably be updated for consistency.

matthias-wright:
Update(f7b1bc5):
- Derive authorized observer keys from the chain bound domain

sebastian-osec:
This looks good to me.

Issue #313, opened 2026-06-01
[Security][tentative] Medium (attackable): Commonware Parent View Not Bound To Summit Parent


evonide:
## Context

Commonware Simplex proposals carry a consensus view and a parent view alongside a payload digest. Summit then fetches the parent block and candidate block by digest and verifies that the decoded Summit block is safe to vote for. The intended invariant is that the Commonware certificate context and the Summit header chain describe the same parent/child relationship.

The sibling candidate-round binding gap is tracked separately. This variant targets the adjacent parent context: the signed Commonware `parent` view is used as a fetch hint, but the verifier does not bind it to the fetched Summit parent block's decoded `view`.

## Claim

A Byzantine proposer submits a Commonware proposal with a chosen `context.parent` view and a payload digest for a Summit child whose header points to a real parent block digest at a different decoded view. Honest validators use the signed parent view only to subscribe/fetch the parent by digest, then `handle_verify` validates the Summit chain using the fetched block's digest and height without rejecting the view mismatch. `handle_verify` accepts a proposal when the fetched parent block's digest matches the child header's `parent`, but it never checks `parent.view() == context.parent.0`. A Byzantine proposer can therefore produce a valid Commonware proposal whose signed parent view differs from the actual Summit parent block view that is used for header continuity and execution checks.

## Flow

The path is on the normal proposal verification flow. It requires an attacker who can propose a block and cause validators to resolve the referenced parent block, and it assumes the Commonware layer permits proposals whose parent view is less than the proposal view even if it is not the decoded view of the Summit parent block. The missing Summit-side binding is in the verifier path that compares the proposal against the fetched parent block.

## Impact

Certificates can attest to one parent-view ancestry context while the executed Summit block chain follows another. That weakens fork-choice and recovery assumptions that downstream certificates, sync requests, and finalized records describe the same header chain that finalizer and application code executed.

## Root Cause

Summit treats the Commonware parent view as a retrieval hint instead of a consensus-bound field that must match the decoded parent block used by the application verifier.

## Code

- The application verifier derives `parent_round` from `context.parent.0` and subscribes by `(round hint, digest)` before fetching the parent: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/application/src/actor.rs#L237-L249
[...]

PR #389, opened 2026-06-10
fix: bind the signed parent view to the fetched parent block during verification

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/313.

Changes:
- `handle_verify` now takes the signed parent view (`context.parent.0`) and rejects a proposal when it disagrees with the fetched parent block's decoded view.
- Exempt the two legitimate cases where they differ:
  - the same-digest boundary re-proposal (returns earlier, so a terminal block re-certified at a later view is unaffected),
  - the first block of an epoch, which carries a signed parent view of 0 (genesis or the previous epoch's terminal block lives in a different view space); the parent is still bound by digest.
- Add regression tests.

### DIFF
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -795,17 +795,36 @@ fn handle_verify<ES: Epocher>(
     }
     // Bind the signed Commonware parent view to the fetched parent block's
     // decoded view so the certificate's ancestry and the Summit header chain
-    // agree. Two cases are exempt:
-    //   - the same-digest boundary re-proposal handled above (already returned),
-    //     where the terminal block is re-certified at a later view;
-    //   - the first block of an epoch, which carries a signed parent view of 0
-    //     (the `parent.0 == 0` sentinel) while its parent is the genesis block or
-    //     the previous epoch's terminal block, in a different view space.
-    // In both cases the parent is still bound by digest (`block.parent()` below).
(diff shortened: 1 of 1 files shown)


sebastian-osec:
I don’t think this fully fixes #313 yet.

The parent-view mismatch check only runs when `signed_parent_view > 0`:

```rust
if signed_parent_view > 0 && parent.view() != signed_parent_view {
    return false;
}
```

That means a proposal with `signed_parent_view = 0` skips the binding entirely. The comment says this is for the first block of an epoch, which makes sense, but the code does not check that the child block is actually the first block of an epoch.

The new `accepts_zero_signed_parent_view_for_first_block` test also looks wrong for that reason. It uses `parent_height = 3` and builds a child at height `4` with `EPOCH_LENGTH = 10`, so this is a mid-epoch block, not an epoch-opening block. The test currently proves that a mid-epoch proposal can use parent view `0` to bypass the binding.

I think the zero-view exception should only apply when the child block is actually the first block of its epoch, for example parent height `9` → child height `10` when the epoch length is `10`. Otherwise a mismatched parent view can still bypass the new check by using `0`.

matthias-wright:
Update(44d171c):
- Constrain parent-view-0 sentinel to genuine epoch openers

sebastian-osec:
Thanks, it looks good to me now.

Issue #343, opened 2026-06-01
[Security][tentative] Medium (attackable): Raw Finalized Header Participant Count Bitmap Allocation


evonide:
## Context

Finalized headers can be imported from checkpoint bundles and served or fetched through RPC/helper paths. Raw SSZ finalized-header decoding uses the header-embedded participant count to configure certificate decoding.

## Claim

A malformed raw finalized header can set a very large `participant_count`. The decoder then uses that unauthenticated value as the certificate participant count, which can drive signer-bitmap allocation before the encoded bitmap bytes have been fully validated.

## Flow

The path is reachable through bootstrap/checkpoint and helper/RPC raw finalized-header decode paths. The scan did not find a live consensus P2P path because syncer finalized/notarized delivery uses epoch-scheme certificate configuration rather than a header-embedded participant count.

## Impact

Checkpoint or finalized-header import can suffer memory pressure or abort before validating the malformed input. Severity is capped below critical because this is a checkpoint/helper/RPC raw-header boundary, not a live consensus P2P path.

## Root Cause

Raw finalized-header decoding uses an unauthenticated embedded participant count as a dependency decoder bound, rather than deriving the certificate participant count from the verified validator set or applying a local maximum before allocation.

## Code

- `Header` raw SSZ fields are decoded before finalized-header certificate handling: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/header.rs#L250-L290

## Related Issues/PRs

The following existing items touch the same trust boundary, adjacent root cause, or likely remediation stack.

- #214
- #235
- #261
- #286
- #314

Note that #343 is raw finalized-header decode allocation from unauthenticated participant count.

## Fix

Do not trust raw `participant_count` as a decoder allocation bound. Cap it to the verified validator set size or a configured maximum before dependency decoding, and add a raw SSZ finalized-header fuzz/unit test with oversized participant counts and short bitmap bytes.

matthias-wright:
Proposed fix: https://github.com/SeismicSystems/summit/pull/390

PR #390, opened 2026-06-10
fix: cap participant count when decoding raw finalized headers

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/343.

Changes:
- Add `MAX_FINALIZED_HEADER_PARTICIPANTS` (100_000) in `types/src/header.rs`.
- Reject a `participant_count` above that cap in `FinalizedHeader::from_ssz_bytes` before the certificate decode (also covers the `Read`/`read_cfg` path, which delegates to it). This bounds the worst-case bitmap allocation to ~12.5 KB while staying far above any real committee size.
- Add a unit test.

### DIFF
--- a/types/src/header.rs
+++ b/types/src/header.rs
@@ -430,6 +430,15 @@ impl fmt::Display for FinalizedHeaderError {
 
 impl std::error::Error for FinalizedHeaderError {}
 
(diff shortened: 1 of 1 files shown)


sebastian-osec:
This looks good to me.

Issue #322, opened 2026-06-01
[Security][tentative] Medium (correctness): Epoch Boundary Restart Keeps Previous View


evonide:
## Context

Commonware Simplex treats views as epoch-local: each epoch has an implicit genesis at view 0 and voting starts at view 1. Summit persists consensus state in the finalizer and uses that persisted `epoch` and `view` to seed syncer's `last_processed_round` on restart.

At an epoch boundary, the finalized terminal block belongs to the old epoch, while the next epoch's early views should still start from view 1. The persisted sync floor must not combine the new epoch number with a view from the previous epoch.

## Claim

A node finalizes an epoch-boundary block at view `V > 1`, persists the post-transition state, and restarts before it has locally cached or finalized early blocks for the new epoch. During new-epoch proposal or verification, application subscriptions for rounds `(epoch + 1, 1..V-1)` can reach syncer; because syncer was seeded with `(epoch + 1, V)`, it drops those missing-block subscriptions as past work. The finalizer increments `state.epoch` at the epoch boundary but does not reset `state.view` before persisting consensus state. Restart then initializes syncer with `(new_epoch, old_terminal_view)`, so syncer can classify valid early new-epoch rounds as older than `last_processed_round`.

## Flow

The path requires an epoch transition finalized at a non-first view, followed by restart from the persisted finalizer DB. The impact requires a missing local block for an early new-epoch proposal or verification subscription; if the needed block is already available locally, the stale round floor is not exercised.

## Impact

Early new-epoch proposal or verification work can be canceled instead of fetched or registered, causing local consensus participation failure after restart. This also gives a concrete producer for the already tracked old-round subscription panic class, but the root cause here is the epoch-boundary persistence mismatch.

## Root Cause

Summit stores one global `view` in consensus state and carries it across an epoch increment even though Commonware views are epoch-local.

## Code

- Finalizer writes the finalized block's view into consensus state: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1415.
- Consensus state stores `epoch` and `view` as global fields: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L21.
- Finalizer loads latest persisted consensus state on startup: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L156.
[...]

PR #391, opened 2026-06-10
fix: reset view at epoch boundary

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/322.

Changes:
- Reset consensus view at epoch boundary.
- Add unit tests.


sebastian-osec:
This fix looks good to me.

Issue #337, opened 2026-06-01
[Security][tentative] Medium (attackable): Observer RPC Reports Master Identity


evonide:
## Context

Observer mode derives a child networking identity from a validator node key so the observer can follow the chain without using the primary validator P2P identity. RPC methods such as `getPublicKeys` and `getDepositSignature` are also identity surfaces for operators and tooling.

## Claim

An operator starts an observer with `--observer <index>` and exposes or uses that observer's RPC endpoint. A client that calls `getPublicKeys` or `getDepositSignature` receives data for the master validator keystore even though the live P2P node is the derived observer key. Observer startup uses the derived child key for P2P, but the normal RPC server receives only the base keystore path. Its public-key and deposit-signature methods reread the master `node_key.pem` and `consensus_key.pem`, so the observer RPC endpoint exposes the validator master identity rather than the active observer transport identity.

## Flow

The path is source-reachable for observer mode. Practical impact depends on whether operator tooling or external clients rely on observer RPC identity methods to create deposits, build allowlists, or reason about which peer identity the endpoint represents.

## Impact

The endpoint can mislead operators and automation into binding or signing for the validator master identity from an observer process. This is lower impact than the general public deposit-signing oracle, but it is a distinct observer-mode identity mismatch.

## Root Cause

Observer key derivation is applied to P2P signing only; RPC identity methods are not given the effective observer public key or an observer-mode policy.

## Code

- Observer mode derives a child signer for P2P configuration: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L375-L385
- Normal RPC startup passes only the base `key_store_path`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L683-L690
- `getPublicKeys` rereads the base node and consensus keys from disk: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/rpc/src/server.rs#L60-L72
- `getDepositSignature` also rereads and signs with the base keystore keys: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/rpc/src/server.rs#L228-L245

## Related Issues/PRs

The following existing items touch the same trust boundary, adjacent root cause, or likely remediation stack.

- #228
- #250
- #283
- #335
- #336

[...]

PR #392, opened 2026-06-11
fix: disable validator-keystore RPC identity surfaces in observer mode

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/337.

Changes:
- Derive the observer child public key in start_network_and_engine and plumb it into the RPC server as observer_node_key: Option<String> (Some only in observer mode).
- getPublicKeys: in observer mode, return the derived child key (the live P2P transport identity) as node and an empty consensus key; validator-mode behavior unchanged.
- getDepositSignature: reject in observer mode with new error DisabledInObserverMode (code 4003) instead of producing master-key deposit signatures.
- Add regression tests.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -750,6 +750,16 @@ where
         authenticated::discovery::Network::new(context.with_label("network"), p2p_cfg);
 
     let oracle = DiscoveryOracle::new(oracle);
+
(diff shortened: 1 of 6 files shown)


sebastian-osec:
This looks good to me.

Issue #336, opened 2026-06-01
[Security][tentative] Medium (correctness): Observer Resolver Can Exclude Its Parent Validator As Self


evonide:
## Context

Summit observer mode derives a child P2P signer from a validator node key. The observer is intended to participate as a secondary peer while the parent validator remains the primary participant for consensus and resolver backfill.

Commonware resolver excludes its own identity from fetch targets. That exclusion is correct only if the resolver identity matches the actual network signer used by the node.

## Claim

An observer is launched with validator master material and `--observer`. If its parent validator is the only reachable, freshest, or targeted resolver peer for a missing block, the observer's resolver can filter that peer out as self and fail or delay backfill. Observer startup derives a child signer for P2P, but the engine initializes the resolver with `self.node_public_key`, the master node key. Commonware resolver then treats the parent validator's primary key as `me`, and can exclude that validator from eligible fetch peers even though it is a distinct reachable node from the observer's effective P2P identity.

## Flow

This affects observer deployments that derive child P2P identity from a validator key. The validated worst case is a partial-connectivity catch-up topology where the parent validator is the only usable resolver source for historical data.

## Impact

Observer catch-up and proof/RPC service can stall or degrade under realistic partial-connectivity conditions. It also weakens recovery assumptions for observer deployments because the effective P2P identity and resolver identity disagree.

## Root Cause

Summit passes the master node public key into the resolver configuration after observer startup has selected a derived child signer for authenticated P2P.

## Code

- Observer startup derives a child P2P signer from the master key: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L375-L383
- Normal validator startup uses the master key directly as the P2P signer: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L400-L407
- Engine resolver configuration uses `self.node_public_key` as resolver public key: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/engine.rs#L438-L442

## Related Issues/PRs

The following existing items touch the same trust boundary, adjacent root cause, or likely remediation stack.

- #228
- #232
- #250
- #335
- #337

Note that #336 is observer self-filtering against parent validator identity.

## Fix

[...]

PR #393, opened 2026-06-11
fix: observer resolver excluding its parent validator as self

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/336.

Changes:
- Add EngineConfig.observer_network_key: Option<PublicKey>: the derived child key, Some only in observer mode
- Set it in start_network_and_engine, deriving the child key once and reusing it for the RPC server's observer identity
- Engine::new resolves a single effective node identity (child key in observer mode, master key otherwise) and uses it for the resolver me, broadcast buffer attribution, and finalizer self-lookup
- Add regression test

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -845,14 +845,16 @@ where
 
     let oracle = DiscoveryOracle::new(oracle);
 
-    // In observer mode the live P2P identity is this derived child key, not
-    // the master node key; the RPC server reports it instead of the keystore
-    // identity and disables keystore-signing methods.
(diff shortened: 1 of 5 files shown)


sebastian-osec:
LGTM. This fixes #336 from the right side of the boundary: observer mode now threads the effective child P2P identity into the engine and resolver instead of letting the resolver use the parent/master validator key.

Optional nit: the observer child key is derived twice in startup, once for the P2P signer and once for `observer_network_key`. Not a blocker, but deriving once and reusing the public key would reduce future drift risk.

Issue #302, opened 2026-06-01
[Security][tentative] Medium (correctness): Application Broadcast Uses A Global Built-Block Slot


evonide:
## Context

Summit's application actor builds block payloads for consensus proposals and later serves Commonware relay requests to broadcast those payloads to peers. Commonware identifies the proposal digest to broadcast and may request targeted forwarding to validators that have not seen the payload.

The intended invariant is that a broadcast callback sends the exact proposal payload requested by Commonware, preserving the digest identity and forwarding plan. Summit's local relay state is therefore part of the boundary between consensus scheduling and block dissemination.

## Claim

An honest leader under delayed or overlapping local proposal and broadcast callbacks can make the application actor drain its single global `built_block` slot for the wrong Commonware digest, so a later block may be broadcast for an earlier request or consumed before its own broadcast while silent voters miss the `Plan::Forward` delivery Simplex requested. The application relay ignores Commonware's requested proposal digest and forwarding plan, then drains one global `built_block` slot, allowing delayed or overlapping broadcasts to send or consume the wrong proposal payload.

## Flow

The path is reachable during normal proposal handling because Summit wires Simplex to the application relay and uses `ForwardingPolicy::SilentVoters`. The wrong-block race requires delayed or overlapping local proposal/broadcast callbacks, which is most plausible under load or around epoch boundaries where overlapping engines can exist; no reviewed evidence shows that a remote Byzantine validator can directly force another node's application actor to drain the slot.

## Impact

A delayed broadcast for proposal A can run after proposal B overwrites the slot. The leader then broadcasts B under A's broadcast callback, or consumes B before its own broadcast. Separately, non-proposer validators that receive a `Plan::Forward` request do not forward the certified proposal to missing voters. Peers that saw consensus traffic for the expected digest may never receive the matching block from the leader or the intended forwarder, causing verification stalls, missed notarization, and avoidable finality delays, especially at epoch boundaries.

## Root Cause

The relay API loses the Commonware broadcast identity: the requested digest and forwarding plan are not used to select or forward the proposal. The application instead stores one mutable `built_block` slot shared across proposals.

## Code

- The relay forwards Commonware's requested digest into `Message::Broadcast`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/ingress.rs#L107.
[...]

PR #394, opened 2026-06-11
fix: implement Plan::Forward so silent voters receive certified blocks

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/302.

Changes:
- Carry `Plan<P>` through `Message::Broadcast` (Message now generic over P)
- Plan::Forward -> syncer.forward(round, payload, peers)
- Plan::Propose -> no-op (block already broadcast at propose time)
- Delete the built_block slot and its stale-broadcast warn
- Require `S: Scheme<Digest, PublicKey = P>` on the application actor
- Unit tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6029,6 +6029,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-engine",
  "anyhow",
+ "commonware-codec",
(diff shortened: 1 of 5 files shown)


sebastian-osec:
No issues here. Looks good to me.

Issue #306, opened 2026-06-01
[Security][tentative] Medium (correctness): Backfill Resolver Uses A Hard-Coded Short Fetch Timeout


evonide:
## Context

Summit uses the syncer resolver to fetch finalized blocks, notarized blocks, and finalization certificates needed by lagging, restarting, or checkpoint-joining nodes. Those backfill requests are part of recovery liveness: if valid historical responses are not accepted, the node can know it is missing finalized data but fail to repair the gap.

Summit already carries a node-level `fetch_timeout` used when configuring the per-epoch Simplex engines. The backfill resolver is wired separately in `Engine::run`, so its timeout must still be long enough for the same P2P and storage environment that serves finalized history.

## Claim

The backfill resolver ignores `EngineConfig::fetch_timeout` and instead hard-codes a 2 second active-request timeout with a 1.5 second retry delay. Commonware removes an active request when that timeout fires, so a valid response that arrives after the 2 second deadline is unmatched and ignored rather than delivered to the syncer.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

Affected nodes can fail to catch up from otherwise valid finalized history even though peers eventually send the requested data. This can become a persistent recovery failure for restarting, joining, or checkpointing validators under slow storage/network conditions, and can compound existing target-pinned or message-size backfill failures.

## Root Cause

Backfill timeout policy is split from Summit's configured fetch timeout. `Engine::run` constructs the resolver with fixed literals instead of passing `cfg.fetch_timeout` or a backfill-specific validated parameter.

## Code

- Summit defines a 5 second `FETCH_TIMEOUT`, stores it in `EngineConfig::fetch_timeout`, and populates that field during config construction: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/config.rs#L22, https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/config.rs#L45, https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/config.rs#L88.
[...]

PR #395, opened 2026-06-12
fix: use configured fetch_timeout for the backfill resolver

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/306.

Changes:
- Add a `fetch_timeout` field to `Engine`, populated from `EngineConfig::fetch_timeout`
- Extract the inline resolver config into `Engine::backfill_resolver_config()`, which uses the configured timeout as the active-request timeout
- Replace the remaining `initial` (1s) and `fetch_retry_timeout` (1.5s) literals with named, documented constants (`BACKFILL_INITIAL_EXPECTED`, `BACKFILL_FETCH_RETRY_TIMEOUT`) — these govern peer-performance priors and retry pacing, not response matching, so they stay independent
- Build the resolver config at the top of `run`, before actors take ownership of engine fields
- Add regression test


sebastian-osec:
This fix looks good to me.

Issue #360, opened 2026-06-01
[Security][tentative] Medium (correctness): Tracked Actor Clean Exits Unsupervised


evonide:
## Context

Summit's engine starts application, buffer, finalizer, syncer, and orchestrator actors as long-lived services. A clean exit from any required actor should normally be treated as a service failure or coordinated shutdown, not ignored until every other actor exits.

## Claim

A consensus-critical actor hits a clean-return path such as syncer application-ack cancellation, syncer metadata sync failure, resolver handler closure, or a closed actor mailbox while sibling actors continue running. The engine does not treat that one clean exit as an immediate failure. `Engine::run` supervises tracked actors with `try_join_all`, which resolves early only on errors. If one actor returns `Ok(())`, the engine continues waiting for the remaining actors, leaving the exited actor's service dead without propagating a failure.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

The node can remain half-alive with a required actor stopped, delaying detection by process supervisors and making P2P/RPC/other consensus components appear alive while a consensus-critical service no longer handles work.

## Root Cause

Long-lived actor handles use normal `Ok(())` return for both coordinated shutdown and local service termination, and `try_join_all` does not supervise "first actor completed" semantics.

## Code

- `Engine::run` waits on tracked actor handles with `try_join_all`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/engine.rs#L467-L475
- Syncer cleanly returns after application ack cancellation: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/syncer/src/actor.rs#L500-L512
- Syncer cleanly returns after application metadata sync failure: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/syncer/src/actor.rs#L523-L527
- Syncer cleanly returns when the resolver handler closes: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/syncer/src/actor.rs#L766-L770
[...]

PR #396, opened 2026-06-12
fix: treat any single tracked-actor exit as an engine failure

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/360.

Changes:
- node/src/engine.rs — replaced try_join_all with first-completion supervision: the actor handles are collected into a named FuturesUnordered, and any completion without a coordinated cancellation cancels the siblings and returns an Err naming the actor, bringing the node down via the supervisor.
- node/src/engine.rs — cancellation is polled first (select_biased!) so a fatal-error self-cancel or committee exit isn't misclassified as an unexpected exit; the graceful-shutdown arm drains all actors and preserves the existing semantics (Ok on clean stop, Err if an actor fails during shutdown).
- node/src/tests/engine.rs (new) — regression test test_engine_detects_single_actor_clean_exit: a failing backfill receiver kills the commonware resolver engine, which closes the handler channel into the syncer, which clean-returns; asserts the engine handle resolves with an error instead of staying pending for 120 simulated seconds. Red on try_join_all, green with this fix.

### DIFF
--- a/node/src/engine.rs
+++ b/node/src/engine.rs
@@ -16,7 +16,7 @@ use commonware_storage::archive::immutable;
 use commonware_utils::acknowledgement::Exact;
 use commonware_utils::{NZU64, NZUsize};
 use futures::FutureExt;
-use futures::future::try_join_all;
+use futures::stream::{FuturesUnordered, StreamExt};
 use governor::clock::Clock as GClock;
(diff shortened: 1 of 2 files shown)


sebastian-osec:
This looks good to me.

Issue #362, opened 2026-06-01
[Security][tentative] Medium (attackable): Withdrawal Overflow Backlog Bypasses The Epoch Cap's Work Bound


evonide:
## Context

Summit limits how many withdrawals are emitted in an epoch with `max_withdrawals_per_epoch`. The finalizer actor also handles consensus-critical mailbox work, including finalized block updates, notarized blocks, height notifications, aux-data requests, and state queries.

## Claim

A normal or adversarially large set of valid withdrawal and exit requests accumulates for one epoch with size far above `max_withdrawals_per_epoch`; at every epoch boundary, the finalizer performs O(backlog) queue collection, overflow rescheduling, and tree rebuild work on the consensus actor path before only a capped prefix can be emitted. The withdrawal cap limits only the withdrawals included in the block, not the finalizer work required to inspect and roll over a larger ready backlog. Epoch-final aux-data collection builds a vector for the full ready epoch before applying `.take(max_withdrawals)`, and epoch transition later reschedules all overflow withdrawals and rebuilds the withdrawal SSZ tree.

## Flow

The path is reachable from accepted withdrawal state when the number of ready withdrawals for an epoch exceeds the cap. The scan did not quantify the backlog size needed for material delay or prove malicious public ingress, so a focused workload still needs to measure finalizer occupancy.

## Impact

The cap can give operators a false resource bound: a large backlog can still stall finalizer mailbox service at epoch boundaries, delaying `SyncerUpdate`, `NotifyAtHeight`, and `GetAuxData` handling. Repeated overflow can carry the same large work set forward into later epochs.

## Root Cause

The code applies the per-epoch cap after collecting the full ready queue and reschedules overflow by moving every remaining key plus rebuilding the complete withdrawal proof subtree.

## Code

- The finalizer actor handles consensus and query mailbox work in one loop: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L306-L383.
- Aux data collects all withdrawals for the epoch before applying `take(max_withdrawals)`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L951-L959.
[...]

PR #397, opened 2026-06-12
fix: bound the deposit-refund backlog by collapsing refunds per address

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/380 (which builds on https://github.com/SeismicSystems/summit/pull/377, https://github.com/SeismicSystems/summit/pull/276, and https://github.com/SeismicSystems/summit/pull/275).

Addresses https://github.com/SeismicSystems/summit/issues/362.

Changes:
- finalizer/src/actor.rs — deposit_refund_key and its wrappers (refunded_deposit_key, invalid_signature_refund_key, malformed_deposit_refund_key, invalid_deposit_tax_key) no longer take a deposit index; key is now (tag, address).
- finalizer/src/actor.rs — removed the now-unused deposit_index param from push_invalid_deposit_withdrawals and updated the three deposit-processing call sites and queue_deposit_refund's key construction. queue_deposit_refund keeps the param (still used in the parse-failure critical-error log).
- finalizer/src/actor.rs — made queue_deposit_refund and DepositRejectionReason pub(crate) for testing.
- finalizer/src/tests/deposit_refunds.rs (new, registered in tests/mod.rs) — many_invalid_deposits_to_one_address_collapse_to_one_entry (50 same-address refunds → one summed entry, one index consumed, incremental root == full rebuild) and distinct_addresses_and_reasons_stay_separate (no over-merging).


sebastian-osec:
I think this is a good partial fix for the deposit-refund amplification case.

The new keying correctly collapses many rejected deposits to the same `(reason tag, withdrawal address)` into one pending refund entry, with the amount summed.

However, I do not think this fully closes #362 as written. The broader invariant is that the withdrawal cap should bound epoch-boundary finalizer work, not only the number of withdrawals emitted in the block.

`get_withdrawals_for_epoch_with_limits` still calls `get_for_epoch_by_kind`, which collects the full per-kind epoch queue into a `Vec` before applying `.take(max_...)`. Epoch rollover also still calls `reschedule_withdrawal_epoch`, which moves all remaining withdrawals and rebuilds the withdrawal SSZ tree.

There is also still an attacker-controlled way to avoid this specific collapse: the merge key includes `withdrawal_address`, and `withdrawal_credentials` comes from the deposit request. So many invalid deposits with many distinct valid Eth1 withdrawal addresses still create many ready refund entries. The second new test confirms distinct addresses remain separate, which is correct for accounting, but means the backlog is bounded by distinct addresses rather than by the configured withdrawal cap.

I think the queue still needs capped-prefix iteration without materializing the full ready epoch, and rollover/SSZ updates need to avoid work proportional to the full overflow backlog.

matthias-wright:
I'm working on a refactor of the deposit and withdrawal processing (out of scope for this audit) that fixes https://github.com/SeismicSystems/summit/issues/362, so I will close this PR.

Issue #334, opened 2026-06-01
[Security][tentative] Medium (attackable): Joining Validators Become Backfill Source Peers


evonide:
## Context

Summit tracks active validators for Simplex consensus, but tracks active plus joining validators in the P2P peer oracle so new validators can connect before they vote. The syncer backfill resolver uses the same peer oracle to choose peers that may serve missing blocks and finalizations to lagging or restarting nodes.

## Claim

A malicious validator submits a valid deposit, reaches `Joining`, and is tracked by honest nodes during the next epoch transition. When a lagging or restarting node requests missing data, Commonware's resolver reconciles its fetch participants from Summit's primary peer set and can send the backfill request to the joining key; that peer can fail to respond, return no data, or force retries until it is blocked. Joining validators are inserted into the primary resolver participant set before Summit proves that they are active voters or serving finalized history. A joining validator that cannot or will not serve backfill data can therefore be selected as a resolver source for lagging nodes.

## Flow

The source handoff is direct. A validly deposited validator can reach `Joining` before activation, a lagging victim can sync from genesis while active serving peers are available, and the resolver can select the joining key for notarized backfill fetches even though that joining peer may not serve archive data reliably.

## Impact

The likely impact is catch-up and recovery delay during validator churn. This does not grant voting power or let a joining validator forge finalized data, so severity should stay below high unless validation shows persistent recovery failure or broad consensus liveness loss from resolver source selection.

## Root Cause

Summit reuses one active-or-joining primary P2P set for both warm-up connectivity and resolver source authorization instead of separating future committee connectivity from peers trusted to serve historical backfill.

## Code

- `get_active_or_joining_validators()` includes `ValidatorStatus::Joining`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/consensus_state.rs#L669-L675
- Epoch transition tracks active-or-joining node keys as primary P2P peers: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L655-L671
- The same transition sends only active validators to the orchestrator for Simplex voting: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L674-L687
[...]

PR #398, opened 2026-06-12
fix: track joining validators as secondary peers, not primary

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/334.

Changes:
- finalizer/src/actor.rs — at the epoch transition, build the peer sets separately: PRIMARY = get_active_validators(); SECONDARY = joining node keys (active-or-joining minus active) + observer keys. Observers are still derived from the full active-or-joining set so joining validators' observer children connect during warm-up.
- finalizer/src/tests/mocks.rs — added RecordingNetworkOracle (and TrackCall) that captures each track() call's epoch index and primary/secondary tiers, since the existing MockNetworkOracle discards them.
- finalizer/src/tests/validator_lifecycle.rs — added test_joining_validator_peer_tier_follows_activation: drives two epoch boundaries and asserts the validator is secondary (not primary) while Joining, then promoted into primary once Active.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -1088,23 +1088,46 @@ impl<
                 self.canonical_state.clear_removed_validators();
             }
 
-            // Create the list of validators for the p2p network for the next epoch.
-            // We also include the validators that already staked and are waiting to join the committee.
-            let active_validators = self.canonical_state.get_active_or_joining_validators();
-            let network_keys: Vec<_> = active_validators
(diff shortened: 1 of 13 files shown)


sebastian-osec:
Looks good to me. Small nit: the test comment is inconsistent with the code. It says the epoch-boundary finalizations happen at heights 5 and 10, but the test attaches finalizations when `height % 5 == 4`, i.e. heights 4 and 9. The comment should say heights 4 and 9.

Issue #333, opened 2026-06-01
[Security][tentative] Medium (correctness): IPv6 Advertised Addresses Use An IPv4-Only Listener


evonide:
## Context

Summit gives Commonware discovery two different addresses during startup: the local socket to bind and the dialable address that peers should learn for this node identity. Operators can source the dialable address from genesis, `--ip`, or public-IP discovery for non-genesis identities. The discovery layer then signs and gossips that dialable address as the peer record other nodes should use.

## Claim

An operator runs a validator, observer, or joining/checkpoint-bootstrapped node with a genesis `ip_address` such as `[2001:db8::10]:18551`, passes `--ip [2001:db8::10]:18551`, or omits `--ip` on a non-genesis identity and receives an IPv6 public-IP response. Startup succeeds, Commonware signs and shares that IPv6 dialable address, but inbound IPv6 dials fail because Summit bound the listener only to `0.0.0.0`. Summit accepts IPv6 dialable addresses from genesis, `--ip`, or public-IP discovery, but both production and local startup hard-code the actual P2P listener to `0.0.0.0:<port>`. A node can therefore advertise a signed IPv6 address while only listening on IPv4.

## Flow

This requires an IPv6 or IPv6-preferred deployment, or an IPv6 result from the fallback public-IP services. The address-family mismatch exists in both production `run_node_inner` and local `run_node_local_inner`; a network integration test should confirm whether outbound-only connectivity masks the impact in some topologies.

## Impact

Peers can learn a valid signed discovery record for an address family the node is not serving. In IPv6-only or IPv6-preferred validator/bootstrap deployments, otherwise syntactically valid Summit configuration can silently produce unreachable peers, isolating validators or delaying recovery/catch-up until operators notice and reconfigure addresses.

## Root Cause

Summit does not derive the listener address family from the configured dialable address and does not reject IPv6 dialable addresses when binding an IPv4-only listener.

## Code

- Genesis validator addresses are parsed directly as `SocketAddr`, which accepts IPv6 socket syntax: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/genesis.rs#L125.
- Explicit `--ip` is also parsed as `SocketAddr`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/args.rs#L770.
[...]

PR #400, opened 2026-06-15
fix: bind P2P listener to the advertised address family

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/333.

Changes:
- Add `wildcard_listen_for(dialable, port)`: binds `0.0.0.0` for an IPv4 advertised address and `[::]` for IPv6 (dual-stack; also accepts IPv4-mapped connections).
- Use it in both `run_node_inner` and `run_node_local_inner` instead of the hard-coded IPv4 wildcard, so the listener family always matches the address Commonware signs and gossips.
- Add unit tests.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -24,7 +24,7 @@ use futures::{FutureExt, channel::oneshot};
 use governor::Quota;
 use ssz::Decode;
 use std::{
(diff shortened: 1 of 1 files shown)


sebastian-osec:
No issues with this one. Looks good to me.

Issue #237, opened 2026-05-22
[Security][tentative] Medium (correctness): Checkpoint Protocol Parameters Are Not Validated Or Re-Derived


evonide:
## Context

Summit's consensus state carries protocol and governance parameters that affect block validation, validator eligibility, timestamp tolerance, deposits, withdrawals, and recovery behavior. During normal operation those values are expected to change only through bounded protocol-parameter paths, while checkpoint and recovery import install serialized consensus state for a joining or recovering node.

## Claim

Checkpoint and recovery import install protocol and governance scalars directly from decoded `ConsensusState` bytes. `verify_checkpoint_chain` does not validate those scalars against normal protocol-parameter bounds or re-derive them from genesis plus signed transition history, so even the documented directory verification path can import consensus-critical policy values that live updates would reject.

An operator or recovery automation starts a node from checkpoint/recovery state whose serialized protocol scalars differ from values accepted through normal genesis or live update paths; checkpoint verification authenticates finalized headers and active membership but not those decoded scalars, so the importer starts under unauthorized policy. Under an operator-trusted checkpoint model this is a startup/recovery correctness failure, not a malicious artifact-supplier attack.

## Flow

The path is reachable when a node starts from a checkpoint or recovered state and runs the checkpoint verifier. The supplied state must preserve the finalized checkpoint artifacts and active validator membership expected by `verify_checkpoint_chain`, while containing serialized protocol fields that the verifier does not inspect.

## Impact

A verified checkpoint or recovered state can carry out-of-bounds or unauthorized policy values: zero or oversized withdrawal caps, oversized deposit caps, out-of-range timestamp windows, altered stake bounds, treasury address changes, or observer caps. Nodes booted from that state can propose or verify blocks under different rules than uninterrupted nodes, causing checkpoint-import divergence, withdrawal starvation, oversized epoch-final blocks, timestamp-window disagreement, or unauthorized governance policy after recovery.

## Root Cause

Checkpoint/recovery decoding treats serialized consensus-state protocol parameters as authoritative state, but the verification path does not apply the protocol-parameter validators or derive those values from authenticated transition history.

## Code

[...]

PR #401, opened 2026-06-15
fix: bound checkpoint/recovery protocol scalars at decode

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/237.

Changes:
- `ConsensusState::read_cfg`: reject `max_deposits_per_epoch` > MAX_MAX_DEPOSITS_PER_EPOCH, `allowed_timestamp_future_ms` outside [MIN, MAX]_ALLOWED_TIMESTAMP_FUTURE_MS, and `observers_per_validator` >
  MAX_OBSERVERS_PER_VALIDATOR — using the same constants as the genesis and runtime paths, mirroring the existing guards.
- Fix `ConsensusState::default()`, which hard-coded `allowed_timestamp_future_ms: 50` (below the protocol minimum of 1_000); set it to MIN_ALLOWED_TIMESTAMP_FUTURE_MS so the invariant holds end-to-end.
- Add negative decode tests for each of the three scalars (valid
  boundaries decode, out-of-range values rejected).

### DIFF
--- a/types/src/consensus_state.rs
+++ b/types/src/consensus_state.rs
@@ -3,7 +3,9 @@ use crate::checkpoint::Checkpoint;
 use crate::dynamic_epocher::DynamicEpocher;
 use crate::execution_request::{DepositRequest, WithdrawalRequest};
 use crate::header::AddedValidator;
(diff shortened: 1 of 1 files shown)


sebastian-osec:
This PR looks good to me.

Issue #215, opened 2026-05-22
[Security][tentative] High (attackable): Checkpoint State Position Is Not Bound To The Verified Header Chain


evonide:
## Context

Checkpoint verification is supposed to let a node trust a compact recovery state by checking a finalized-header chain and the checkpoint data it authenticates. The decoded `ConsensusState` carries the node's consensus position, forkchoice, head digest, and epoch schedule, which should describe the same point in history as the verified terminal header.

## Claim

`verify_checkpoint_chain` verifies finalized headers and the terminal checkpoint-data hash, but does not bind the decoded `ConsensusState` position to that terminal header chain. The checkpoint bytes can carry independent `epoch`, `view`, `latest_height`, `head_digest`, forkchoice, and dynamic-epocher state.

A malicious checkpoint provider can give a node a checkpoint whose terminal finalized header already signs the supplied checkpoint bytes, while the decoded state inside those bytes advertises a different consensus position or epoch schedule. A passive public distributor cannot mutate an honestly signed checkpoint into this form because verification recomputes `sha256(checkpoint.data)` and compares it to the signed terminal header.

## Flow

The path is reachable during checkpoint startup from a verified checkpoint only when the supplied finalized-header chain and terminal checkpoint digest are otherwise accepted. That means the altered decoded fields must be inside bytes already authenticated by the terminal header, or the node must be using an unauthenticated single-file checkpoint path covered by the separate startup-verification finding. No evidence was found for a public RPC/P2P actor or passive mirror forcing altered state-position bytes into a verified checkpoint.

## Impact

A checkpoint can pass trustless header verification while its decoded state claims a height, epoch, head digest, forkchoice, or epoch schedule that does not correspond to the verified terminal header. An importing node can then start from the wrong consensus position and use incoherent state when serving RPC, syncing forward, or classifying future epoch boundaries.

The practical impact is wrong-state startup from a selected checkpoint artifact. Depending on which decoded fields are inconsistent, the node can stop during later application paths or continue with checkpoint-derived consensus metadata that was not bound to the verified header chain.

## Root Cause

Checkpoint verification authenticates the checkpoint data hash but does not derive or validate decoded state position, forkchoice, or dynamic-epocher horizon against the terminal finalized header.

## Code

- Checkpoint header verification checks epoch contiguity while walking finalized headers: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L218.
[...]

PR #402, opened 2026-06-16
fix: bind checkpoint state position to the verified terminal header

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/215.

Changes:
- Add a position-binding step to verify_checkpoint_chain. Since the checkpoint is created at the penultimate block of an epoch and the terminal header is the last block, an honest checkpoint satisfies: latest_height + 1 == terminal.height(), head_digest == terminal.parent(), epoch == terminal.epoch(). Reject any checkpoint that doesn't.
- Add CheckpointVerificationError::CheckpointStatePositionMismatch.
- Parameterize checkpoint_verification_fixture by (state_latest_height, terminal_header_height) so it models the penultimate→last structure.
- Add regression test test_checkpoint_verifier_binds_state_position_to_terminal_header: a consistent (5,6) checkpoint verifies; an inconsistent (5,9) one is rejected.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -658,7 +658,7 @@ impl<
                                 self.db
                                     .get_finalized_header(epoch - 1)
                                     .await
-                                    .map(|h| h.header.digest.0)
+                                    .map(|h| h.header().get_digest().0)
                             } else {
                                 // Future epoch we have not reached: no genesis to serve.
                                 None
--- a/node/src/test_harness/common.rs
+++ b/node/src/test_harness/common.rs
@@ -283,7 +283,7 @@ pub fn run_until_height(
         }
 
         // Verify all validators share the same state root
-        assert_state_root_consensus(&consensus_state_queries).await;
+        assert_state_root_consensus_synced(&context, &consensus_state_queries, &[]).await;
 
         context.auditor().state()
     })
--- a/node/src/tests/checkpointing/creation.rs
+++ b/node/src/tests/checkpointing/creation.rs
@@ -213,7 +213,7 @@ fn test_checkpoint_created() {
                 .is_ok()
         );
 
-        common::assert_state_root_consensus(&consensus_state_queries).await;
+        common::assert_state_root_consensus_synced(&context, &consensus_state_queries, &[]).await;
 
         context.auditor().state()
     });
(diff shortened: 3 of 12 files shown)


sebastian-osec:
I think this is a good partial fix for the positional part of #215. The new checks correctly bind the decoded checkpoint state to the verified terminal header for:

- `latest_height + 1 == terminal.height()`
- `epoch == terminal.epoch()`
- `head_digest == terminal.parent()`

That matches the honest checkpoint flow, where the checkpoint is created from the penultimate block and committed by the terminal block.

However, I do not think this fully fixes #215 as written. The issue also calls out other decoded `ConsensusState` fields that are still independent checkpoint contents, especially `forkchoice`, `epoch_genesis_hash`, and the embedded `DynamicEpocher`.

Those fields are still controlled by whoever constructs the checkpoint artifact. A malicious checkpoint can satisfy the new height/epoch/head checks while still choosing a forged `forkchoice` or inconsistent epoch schedule, as long as those checkpoint bytes are committed by the terminal header. That matters because startup reads `canonical_state.get_forkchoice()` and commits that execution hash.

The added test covers the `(latest_height, terminal_height)` mismatch with `(5, 9)`, but it does not test mutated `forkchoice`, `epoch_genesis_hash`, or an invalid `DynamicEpocher` horizon/schedule.

Small defensive note: `saturating_add(1)` would be better as `checked_add(1)` so an overflowed `latest_height` cannot accidentally satisfy the invariant.

matthias-wright:
https://github.com/SeismicSystems/summit/pull/422 (which builds on this PR) prevents loading a single-checkpoint file without verification (unless an unsafe flag is set), and also adds verification, such as checking that `DynamicEpocher` actually covers the decoded epoch/latest_height.
I think https://github.com/SeismicSystems/summit/pull/422 fixes https://github.com/SeismicSystems/summit/issues/215, so we should be able to close this PR once https://github.com/SeismicSystems/summit/pull/422 is merged.

sebastian-osec:
That makes sense to me. Since #422 builds on this branch and includes the checkpoint-position binding commit from #402, I think #402 can be closed as superseded once #422 is merged.

#422 also covers the follow-up concerns here by requiring verified checkpoint startup by default, binding `last_block` to the verified terminal finalization payload, using the verified terminal header instead of the separately loaded artifact, and checking that the embedded `DynamicEpocher` covers the decoded epoch/latest_height.

matthias-wright:
Superseded by https://github.com/SeismicSystems/summit/pull/422.

Issue #340, opened 2026-06-01
[Security][tentative] Medium (attackable): Permissioned Pause Signatures Are Replayable


evonide:
## Context

Permissioned pause exposes signed RPC actions that let an authorized administrator pause or unpause local consensus participation. The signature is the authorization boundary for a control-plane action that directly changes whether the validator proposes and verifies.

Each accepted authorization should be bound to its intended deployment and request instance, so obtaining one valid request does not grant reusable pause control.

## Claim

Anyone who obtains a valid `pause` or `unpause` request can replay the same signature during the accepted freshness window. If the same admin key is reused, the signature may also authorize the same action on another Summit deployment. Permissioned pause violates the authorization uniqueness invariant by signing only an action and timestamp. Because the message omits deployment scope, target scope, nonce, and monotonic sequence, a valid pause or unpause signature remains replayable within the freshness window and can also apply across deployments that share the admin key.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

Anyone who obtains a valid pause or unpause signature can replay it for the accepted freshness window. Because the signature is not chain- or node-scoped, the same signature can apply to other Summit deployments using the same admin key if their permissioned RPC is reachable. This can stall consensus participation or race an intended unpause.

## Root Cause

The signed authorization message contains only the action and timestamp, with no deployment scope, target scope, nonce, or monotonic replay barrier.

## Code

- Pause auth signs messages of the form `summit-pause-v1:{action}:{timestamp}`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/auth.rs#L7, https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/auth.rs#L10.
- Verification checks only that fixed message, timestamp freshness, and admin address: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/auth.rs#L52.
- `pause` / `unpause` flip the shared consensus pause flag after that check: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L381.
[...]

PR #404, opened 2026-06-16
fix: bind pause/unpause RPC authorization to the deployment

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/340.

Changes:
- types: add `pause_signature_domain(genesis_hash, namespace)`, sharing a new `signature_domain(tag, …)` helper with `deposit_signature_domain` (distinct `summit-pause-v1` tag; deposit domain bytes unchanged).
- rpc/auth: signed message is now `summit-pause-v1:{scope_hex}:{action}:{timestamp}`; `verify_action` takes the deployment scope.
- rpc/server: `SummitRpcServer` stores `pause_scope` (hex of the pause domain), computed in `new()` from the genesis hash + namespace it already receives.
- rpc/auth tests: thread the scope through existing tests; add `rejects_other_deployment_scope` (a signature for net-a is rejected on net-b).

Note:
I don't think a nonce is necessary, since the signed pause message contains a timestamp.
The timestamp allows replays only in the freshness window.
Since the pause is idempoitent, even a replay within the freshness window will not affect the network.
The only realistic scenario is if the network is paused and unpaused inside of the freshness window of the pause message..
An attacker could replay the pause message to undo the unpause. However, this can be countered by unpausing the network again.
It is also likely, that the intended network pause will outlast the freshness window of the pause message.
Cross-deployment replays are an issue, so this PR adds the namespace and genesis hash to the signed message.

### DIFF
--- a/rpc/src/auth.rs
+++ b/rpc/src/auth.rs
@@ -25,18 +25,32 @@ fn now_secs() -> Result<u64, RpcError> {
         .map_err(|e| RpcError::Internal(format!("system clock before unix epoch: {e}")))
 }
 
+/// `scope` is the hex of the deployment-bound pause domain
+/// (`summit_types::pause_signature_domain` over the EL genesis hash +
+/// namespace). Binding it into the signed message scopes the authorization to
+/// this deployment, so a pause/unpause signature cannot be replayed against
+/// another network that trusts the same admin key.
 pub fn verify_action(
+    scope: &str,
     action: &str,
     timestamp_secs: u64,
     signature_hex: &str,
 ) -> Result<(), RpcError> {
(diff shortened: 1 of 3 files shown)


sebastian-osec:
That makes sense. This PR fixes the cross-deployment replay issue by binding the signed message to the genesis hash + namespace.

For same-deployment replay, the signature can still be reused within the timestamp window, but I agree the impact is narrower: pause is mostly idempotent, and the main bad case is an old pause replayed after a quick unpause inside the same window. If we’re accepting that as an operational tradeoff instead of adding nonce/sequence tracking, this looks good to me.

Issue #329, opened 2026-06-01
[Security][tentative] Medium (correctness): Fresh Finalization Can Bind To A Stale Same-Height Archived Block


evonide:
## Context

Summit's syncer stores finalized blocks and epoch-boundary finalization certificates in separate immutable archives keyed by height. A lagging node can later ask peers for a `Request::Finalized { height }` response, then dispatches the next archived block and, for epoch-boundary heights, independently loads the archived finalization for the same height before reporting the pair to the finalizer.

## Claim

A node reaches a state where its finalized-block archive contains a stale block at an epoch-boundary height but its finalization archive lacks or later accepts a finalization for that height, for example after stale local archive reuse, interrupted repair, checkpoint/restart composition, or another already-tracked archive inconsistency. A valid peer response for `Request::Finalized { height }` then supplies the correct block and certificate, but the duplicate-index immutable block archive keeps the old block while the finalization archive accepts the certificate, letting subsequent dispatch bind them together. When a stale finalized block already exists at height `H`, a later valid `Request::Finalized { height: H }` can store the fresh finalization while the immutable finalized-block archive ignores the replacement block. Dispatch then pairs the stale archived block with the fresh finalization by height, and the finalizer only checks `block.header.digest == finalization.proposal.payload` with a release-disabled `debug_assert!`.

## Flow

The source path is reachable through syncer finalized delivery and dispatch. The validated sequence is distinct from pure local archive tampering because the second half of the chain uses a fresh peer-supplied finalization response that should repair the height but cannot replace the existing same-height block.

## Impact

At an epoch boundary, Summit can execute or keep the stale block as canonical while constructing a finalized header that carries a finalization certificate for a different digest. If accepted by a storage backend or replay path, that mismatch can poison finalized-header exports, checkpoint verification material, recovery state, and downstream operators that trust the stored header/certificate bundle.

## Root Cause

The syncer treats finalized blocks and finalizations as independently replaceable height-indexed records even though Commonware's immutable archive silently ignores duplicate block indexes, and the finalizer does not release-enforce the block/certificate digest binding before storing finalized headers.

## Code

- Summit pins `commonware-storage = "2026.4.0"` in the target source: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/Cargo.toml#L20.
[...]

PR #405, opened 2026-06-16
fix: enforce finalized block/certificate binding

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/329.

Changes:
- syncer store_finalization: reject the write if a different block already occupies the height (immutable archive can't replace, so fail-stop rather than form a half-updated stale-block/fresh-cert pair).
- syncer try_dispatch_blocks: verify stored block.digest() == finalization.proposal.payload before reporting a finalized block at an epoch boundary; halt dispatch on mismatch.
- finalizer: replace the epoch-boundary debug_assert! with a release-enforced check that fail-stops (cancels the node) before storing a finalized header.
- test: test_finalizer_rejects_block_certificate_digest_mismatch — a boundary block paired with a certificate over a different digest must cancel the finalizer instead of storing a misbound header.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -1193,13 +1193,31 @@ impl<
             // the finalization.
             let finalization = finalization
                 .expect("finalization is always included for the last block of an epoch");
-            debug_assert!(block.header.get_digest() == finalization.proposal.payload);
+            // Release-enforced binding: never construct a finalized header from a
+            // block whose digest the certificate does not finalize. The syncer
+            // pairs the block and certificate by height from two independently
+            // keyed archives; a mismatch here means an upstream archive
+            // inconsistency, and exporting it would poison finalized-header and
+            // checkpoint material. Fail-stop instead of trusting it.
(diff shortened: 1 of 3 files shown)


sebastian-osec:
This looks like it fixes the main #329 path in the syncer: `store_finalization` now refuses to store a finalization if a different block already exists at that height, and dispatch also checks `finalization.proposal.payload == block.digest()` before reporting the pair.

One small concern: the finalizer fallback check happens after the block has already been executed, forkchoice committed, finalized height notified, and the syncer acked. Since the check is meant to prevent a block/certificate mismatch, should it run before `execute_block(...)` instead?

The normal syncer path should now prevent this case, so I don’t think this is the original issue still being open. But as defense-in-depth, the finalizer should probably reject the mismatched pair before applying any state changes.

matthias-wright:
Update(5f30f5d):
- Bind finalized block to certificate before executing it

sebastian-osec:
This looks good to me now

Issue #355, opened 2026-06-01
[Security][tentative] Medium (attackable): State-Proof Pubkey Parsing Allocates Oversized Hex Before Length Checks


evonide:
## Context

Summit's proof RPC lets external clients request Merkle proofs for scalar state fields, validator accounts, validator fields, and withdrawal records. The RPC handler parses caller-supplied key descriptors before sending valid proof requests to the finalizer actor, so malformed descriptors should be rejected cheaply at the public RPC boundary.

The trust boundary is the unauthenticated JSON-RPC listener. Request-size caps limit the body, but parsing should still validate fixed-size key encodings before allocating decoded buffers.

## Claim

A remote caller with access to the public RPC port sends `getStateProof` with one or more pubkey descriptors containing a large even-length hex string that is not exactly 64 hex characters. Summit accepts up to a 10 MiB request body on the normal RPC server, then performs the avoidable hex decode and allocation inside the RPC handler before returning an invalid-key error. `getStateProof` parses every caller-supplied descriptor with `parse_key`, and pubkey-bearing descriptors call `parse_hex_pubkey`. That helper strips an optional `0x` prefix, calls `hex::decode` on the entire remaining string, and only then checks whether the decoded bytes are exactly 32 bytes. Because Alloy's `hex` dependency is `const-hex`, `decode` allocates an output vector sized to half the input before decoding, so a malformed-but-large `validator:`, `validator_field:`, `withdrawal:`, or `withdrawal_field:` key can force oversized allocation and decode work before Summit rejects it.

## Flow

The normal RPC server merges the proof API into the public listener, configures 1000 connections, and allows 10 MiB request bodies. The attacker does not need valid chain state or finalizer access because the malformed descriptor is rejected before `generate_state_proof` is called. Practical impact depends on concurrent request volume and allocator behavior under near-cap request bodies.

## Impact

The issue gives a public RPC caller a cheap request-side memory and CPU amplification path. The strongest source-backed impact is targeted RPC availability degradation on exposed nodes; it does not prove consensus-wide liveness failure because malformed keys fail before finalizer mailbox work.

## Root Cause

Fixed-length pubkey descriptors are decoded into a heap `Vec<u8>` before Summit validates that the hex payload has the only acceptable encoded length.

## Code

[...]

PR #406, opened 2026-06-16
fix: validate proof pubkey length before decoding

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/355.

Changes:
- parse_hex_pubkey: strip optional 0x, reject in O(1) (no allocation) if the remaining string isn't exactly 64 chars, then decode into a fixed [u8; 32] via hex::decode_to_slice. Covers all four pubkey descriptors via the shared helper.
- test parse_pubkey_rejects_wrong_length_before_decoding: an oversized (8192-char) and a too-short hex key are rejected; the valid 64-char form still parses.


sebastian-osec:
This looks good to me. No issues.

Issue #349, opened 2026-06-01
[Security][tentative] Medium (attackable): RPC WebSocket Connections Can Exhaust Server Permits


evonide:
## Context

Summit's RPC wrapper builds both the normal public RPC server and the temporary first-boot genesis RPC server on top of jsonrpsee. The normal RPC server sets method body/response limits and a connection count, while the genesis RPC uses the builder defaults.

RPC availability is an operator-facing dependency for provisioning and monitoring. Connection admission therefore needs its own bounds, independent of method-level finalizer work.

## Claim

A remote host that can reach the normal RPC port opens and holds roughly the configured 1000 WebSocket connections, or reaches the temporary genesis RPC and holds its default jsonrpsee connection permits. Because idle WebSocket tasks retain their permits, subsequent legitimate RPC calls are rejected or cannot connect; on first boot this can block the legitimate `sendGenesis` provisioning call while the node waits for genesis. Summit leaves jsonrpsee WebSocket handling enabled with no ping or idle-connection eviction configured. A remote client can open idle WebSocket connections until the server's connection permit limit is exhausted, denying later HTTP or WebSocket requests without invoking any RPC method.

## Flow

The normal RPC listener binds to all interfaces and sets `max_connections(1000)`. The genesis RPC listener also binds through the same builder and does not override jsonrpsee's default connection count. jsonrpsee 0.26 enables WebSockets and disables WebSocket pings by default, so upgraded idle clients retain connection permits until client close, transport error, or server shutdown.

## Impact

An exposed node can lose its operational RPC surface to cheap idle connections. The first-boot variant can deny the genesis write path and keep the node stuck in provisioning even before any unauthenticated genesis overwrite occurs.

## Root Cause

The RPC builder configures request and response body caps but does not disable WebSocket upgrades, configure WebSocket ping/idle timeouts, or add a lower admission policy for the genesis server.

## Code

- `RpcServerBuilder::new` binds `0.0.0.0` and starts from `ServerConfigBuilder::new()` without WebSocket-specific settings: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/rpc/src/builder.rs#L36-L40
- The builder passes that config directly into jsonrpsee: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/rpc/src/builder.rs#L73-L76
[...]

PR #407, opened 2026-06-16
fix: disable WebSocket upgrades on RPC servers

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/349.

Changes:
- RpcServerBuilder::new and new_localhost: build config from ServerConfigBuilder::new().http_only(), disabling WS upgrades for all ervers (public, admin, genesis) at the single construction point.
- test test_websocket_upgrades_are_rejected: HTTP health still works; a ws:// connect returns Err.


sebastian-osec:
LGTM.

Issue #347, opened 2026-06-01
[Security][tentative] Medium (attackable): HTTP Slow-Body Requests Can Exhaust RPC Permits Before Method Dispatch


evonide:
## Context

Summit exposes normal and first-boot genesis JSON-RPC listeners through jsonrpsee. The RPC layer configures connection and body-size limits, but those limits need to account for transport-level clients that consume admission capacity before a JSON-RPC method is parsed or dispatched.

The trust boundary is the public HTTP listener. Method-level cost limits, finalizer-mailbox isolation, and response caps do not help if an attacker can hold every server permit while the HTTP body is still being read.

## Claim

A remote client with network access to the normal RPC port or temporary genesis RPC port opens enough HTTP POST requests with JSON content and sends only partial bodies, or sends body frames slowly enough to keep `read_body` pending. Each pending request holds a jsonrpsee permit, and legitimate RPC calls are rejected or blocked until the slow bodies close. jsonrpsee acquires a connection permit before it reads the HTTP request body and holds that permit until `call_with_service` finishes. Summit does not configure a request-body read deadline, so a reachable caller can open many JSON-looking POST requests and drip or withhold body frames until all RPC permits are occupied before any method dispatch occurs.

## Flow

Reachable when the RPC listener is exposed, which is the default bind address in Summit's builder. Normal RPC sets `max_connections(1000)`, 10 MiB body, and 10 MiB response caps; genesis RPC uses the same builder with wildcard CORS and the default connection configuration. The practical number of sockets and timing impact still need runtime validation.

## Impact

An attacker can deny public RPC access for an affected node, including first-boot genesis provisioning, before any Summit method handler or finalizer mailbox work is reached. This is a targeted node availability issue rather than a proven consensus-wide liveness failure.

## Root Cause

Summit relies on jsonrpsee connection permits and body-size caps without setting an HTTP request-body read timeout, slow-client eviction, or front-end rate control.

## Code

- Summit's RPC builder binds to all interfaces by default: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/rpc/src/builder.rs#L36.
[...]

PR #408, opened 2026-06-16
fix: bound RPC request time via configurable timeout

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/347.

Changes:
- rpc builder: always apply tower_http::timeout::TimeoutLayer; add a with_request_timeout setter (enable tower-http "timeout" feature).
- RpcBodyLimits: add request_timeout; thread it through the public, admin, and genesis start paths.
- CLI: new --rpc-request-timeout-secs flag (default 30s), wired into RpcBodyLimits; add the field to the e2e/testnet RunFlags literals.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6714,6 +6714,7 @@ dependencies = [
  "http-body",
  "iri-string",
  "pin-project-lite",
+ "tokio",
  "tower 0.5.2",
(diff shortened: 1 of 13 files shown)


sebastian-osec:
Looks good to me.

Issue #348, opened 2026-06-01
[Security][tentative] Medium (attackable): JSON-RPC Batch Amplification


evonide:
## Context

The RPC builder does not configure a JSON-RPC batch limit, so a single request can contain many calls up to the body-size limit.

## Claim

A remote caller reaches the public RPC port and submits large JSON-RPC batches containing expensive methods such as checkpoint queries, public-key reads, deposit-signature generation, or state-proof requests. The RPC builder does not configure a JSON-RPC batch limit, so a single request can contain many calls up to the body-size limit.

## Flow

Confirmed reachable on the public node RPC path. `run_node_inner` starts `start_rpc_server` from the configured RPC port, `RpcServerBuilder::new` binds to `0.0.0.0`, and the server config sets connection/body/response limits without setting jsonrpsee's batch request config. The remaining prerequisite is ordinary network reachability to that public RPC listener.

## Impact

Batch amplification can increase CPU, disk I/O, serialization, and finalizer-mailbox pressure beyond what per-request body limits imply. The strongest source-backed impact is availability degradation of public RPC and additional finalizer actor occupancy, not proven network-wide consensus failure.

## Root Cause

Connection and body limits are configured without a corresponding RPC batch-size or method-cost limit.

## Code

- Node startup launches the public RPC task with the configured `rpc_port`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L683.
- RPC server registers public, proof, and permissioned method sets, then sets `max_connections(1000)`, a 10 MiB request body cap, and a 10 MiB response cap without a batch cap: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/lib.rs#L47.
- The builder binds on `0.0.0.0`, exposes setters for connection/request/response/CORS limits only, and passes the built config directly to jsonrpsee: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/builder.rs#L35.
- Public method names include checkpoint, finalized header, deposit-signature, epoch-bounds, and state-proof endpoints: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/api.rs#L17.
- `getCheckpoint`/`getLatestCheckpoint` query the finalizer and serialize checkpoint, last-block, and finalized-header SSZ bytes: https://github.com/SeismicSystems/summit/blob/ed2c5c8/rpc/src/server.rs#L74.
[...]

PR #409, opened 2026-06-16
fix: cap JSON-RPC batch size

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/408.

Addresses https://github.com/SeismicSystems/summit/issues/348.

Changes:
- rpc builder: default to BatchRequestConfig::Limit(DEFAULT_RPC_MAX_BATCH_SIZE) in new/new_localhost; add a with_batch_limit setter (0 disables batching, n caps the batch).
- RpcBodyLimits: add max_batch_size; thread it through the public, admin, and genesis start paths.
- CLI: new --rpc-max-batch-size flag (default 50, 0 = disabled), wired into RpcBodyLimits; add the field to the e2e/testnet RunFlags literals.
- test test_oversized_batch_is_rejected: a within-limit batch is served, an over-limit batch is rejected.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -12,8 +12,8 @@ use commonware_cryptography::{Signer, certificate::Scheme};
 use commonware_p2p::{Ingress, authenticated};
 use commonware_runtime::{Handle, Metrics as _, Runner, Spawner, tokio};
 use summit_rpc::{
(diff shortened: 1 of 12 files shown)


sebastian-osec:
This looks good to me.

Optional test hardening: it may be worth adding a small regression test for `max_batch_size = 0` to prove batching is actually disabled, and one custom-limit test to prove the CLI/configured value is honored.

Issue #363, opened 2026-06-01
[Security][tentative] Medium (attackable): Withdrawal Queue Root Omits Executable Schedule


evonide:
## Context

`WithdrawalQueue` stores pending withdrawals redundantly: a map from pubkey to withdrawal data and a per-epoch schedule that decides when each withdrawal becomes executable. The SSZ root is expected to commit to the restored queue state that block aux data, finalization, and proof consumers use.

The executable schedule is not just an index cache. It determines which map entries are emitted for a given epoch and how pending withdrawals are ordered.

## Claim

A malformed checkpoint or persisted-state source can encode a withdrawal map and schedule that decode successfully but disagree; the SSZ rebuild follows only scheduled epoch slots and filtered `get_for_epoch` values, so schedule keys or unscheduled map entries can escape the root commitment that proof consumers see. The withdrawal SSZ root is rebuilt by iterating schedule-reachable withdrawals, but it does not independently commit the schedule epoch keys or map entries that are not reachable through the schedule. A decoded queue can therefore contain live withdrawal map or schedule state that affects later behavior while being omitted or under-bound in the rebuilt root/proof structure.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

An imported node can hold withdrawal state that is live for map-based queries or later merge logic but absent from the SSZ commitment, or can execute withdrawals under schedule keys that are not themselves authenticated by the root. This weakens state-root and proof soundness for withdrawal accounting around checkpoint/recovery boundaries.

## Root Cause

`rebuild_withdrawals` treats the schedule as an internal positional index and commits only the filtered withdrawal values, not the full redundant schedule/map relation.

## Code

- `WithdrawalQueue` serializes the withdrawal map and schedule independently: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/withdrawal.rs#L341.
- Deserialization returns both structures without cross-validation: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/withdrawal.rs#L364.
- The schedule controls `get_for_epoch` and filters scheduled pubkeys through the map: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/withdrawal.rs#L236.
[...]

PR #410, opened 2026-06-16
fix: cross-validate withdrawal map and schedule at decode

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/363.

Changes:
- WithdrawalQueue::read_cfg: after decoding, reject unless every scheduled pubkey is in the map under the same epoch, scheduled exactly once, and every map entry is scheduled. Makes map/schedule consistent so the existing root soundly reflects the queue.
- tests (added test-first): decode_accepts_consistent_queue plus decode_rejects_{unscheduled_map_entry, scheduled_pubkey_absent_from_map, schedule_epoch_mismatch}.

### DIFF
--- a/types/src/withdrawal.rs
+++ b/types/src/withdrawal.rs
@@ -400,6 +400,46 @@ impl Read for WithdrawalQueue {
             schedule.insert(epoch, pubkeys);
         }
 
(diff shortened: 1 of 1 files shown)


sebastian-osec:
Looks good to me.

Issue #324, opened 2026-06-01
[Security][tentative] Medium (attackable): Finalized Sync Does Not Enforce Parent Linkage


evonide:
## Context

Finalized sync lets a lagging or repairing node import certified finalized blocks from peers or archived history without going through live proposal verification. The syncer stores and dispatches those blocks to application and finalizer components so the node's canonical finalized state can catch up by height.

Because this path accepts externally supplied finalized history, each imported block must still extend the node's current finalized head. A certificate for a block at height `H` proves finality for that block, but the local canonical chain remains coherent only if the block's parent matches the stored finalized block at `H-1`.

## Claim

A corrupted local archive, stale cross-context store, or same-namespace/key foreign sync source that already possesses a valid finalization certificate for a block at height `H` whose Summit parent is not the node's finalized block at `H-1` can trigger the missing parent check; the syncer accepts and dispatches by height/certificate alone, allowing the node to advance canonical finalized state onto a wrong-parent history. A normal same-chain peer still cannot forge the required finalization certificate for an arbitrary wrong-parent block, so the malicious path depends on an existing valid certificate source such as the `live-consensus-domain-not-chain-bound` cross-deployment replay condition. The finalized catch-up/backfill path accepts finalized blocks by height and certificate payload without enforcing that `block.parent()` matches the local finalized head. A certified finalized block can therefore be stored, dispatched, and executed as canonical even when it does not extend the node's current finalized chain.

## Flow

The code path is reachable during finalized sync, repair, or startup dispatch when blocks arrive through the syncer finalized delivery path rather than live proposal verification. The bad state additionally requires a finalization certificate that verifies for the delivered block digest while the node's local height `H-1` finalized block is not that block's parent. The concrete adversarial handoff is a peer or archive from another Summit deployment that reuses the live namespace and validator keys: `live-consensus-domain-not-chain-bound` lets that foreign finalization verify locally, and this missing parent check lets the wrong-parent block pass into finalized dispatch.

## Impact

A lagging or repairing node can advance canonical CL state to an impossible history, poisoning finalized headers, checkpoints, forkchoice metadata, and later proposal or verification state. Even if the execution layer rejects the foreign payload against the local EL head, `execute_block` keeps the block in the CL chain and updates canonical height/view/head digest, so the divergence is a consensus-state effect rather than only a payload-execution effect.

## Root Cause

[...]

PR #411, opened 2026-06-16
fix: enforce parent linkage for catch-up finalized blocks

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/324.

Changes:
- finalizer handle_finalized_block: in the catch-up direct-execute branch, reject (critical error + fail-stop) a finalized block whose parent does not equal the canonical head digest, before executing it.
- test (added test-first) test_finalizer_rejects_finalized_block_with_wrong_parent: a correctly-chained sequence is accepted; a height-H block with the wrong parent fail-stops the finalizer.


sebastian-osec:
Looks good to me. No issues.

Issue #342, opened 2026-06-01
[Security][tentative] Medium (correctness): Protocol Parameter Queue Rebuilds SSZ Subtree Per Accepted Record


evonide:
## Context

The intended resource invariant is that a valid batch of parameter changes should scale roughly linearly with the number of records, or be capped before it can monopolize the finalizer actor.

## Claim

An authorized protocol-parameter source, compromised parameter owner, or execution-layer source that can emit many valid type-`0xFF` records in an accepted payload causes every validator to process the batch. The finalizer appends and rebuilds the subtree for each record before returning to other finalizer mailbox work. Grouped type-`0xFF` protocol-parameter entries are parsed until the input buffer is empty, and the finalizer pushes each accepted record one by one. Each push immediately rebuilds the complete pending protocol-parameter SSZ subtree, so `N` accepted records cause roughly `1 + 2 + ... + N` rebuild work and repeated tree allocations inside finalizer block execution.

## Flow

The source path is live for accepted execution requests, but production reachability inherits the existing protocol-parameter authority caveat: normal deployments may gate emission through an owner-controlled contract. This scan did not quantify how many records are needed for observable finalizer delay.

## Impact

A large accepted parameter batch can stall block execution, finalizer mailbox service, and checkpoint/export work. If the batch lands near an epoch boundary, the enlarged pending queue can also be serialized into checkpoint state before cleanup.

## Root Cause

`push_protocol_param_change` rebuilds a collection subtree for every single append instead of batching the queue mutation or enforcing a per-block/per-epoch count cap before rebuild.

## Code

- Grouped protocol-parameter parsing loops until the buffer is empty: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/execution_request.rs#L281-L291
- The finalizer handles every decoded protocol-parameter request and calls `push_protocol_param_change`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L1696-L1703
- `push_protocol_param_change` appends one item and immediately rebuilds the subtree from the whole queue: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/consensus_state.rs#L343-L346
[...]

PR #412, opened 2026-06-16
fix: batch protocol-param SSZ rebuild in finalizer

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/342.

Changes:
- types/consensus_state.rs: add push_protocol_param_changes(impl IntoIterator) which extends the queue then rebuilds the param subtree at most once (no-op on an empty batch)
- finalizer/actor.rs: process_execution_requests now accumulates accepted ProtocolParams into a batch across the whole block and flushes once after the request loop instead of rebuilding per record, making the per-block rebuild O(N)
- tests: add test_batch_protocol_param_changes_match_per_record (batch root == per-record root == full rebuild) and test_empty_protocol_param_batch_is_noop


sebastian-osec:
This looks good to me.

Issue #317, opened 2026-06-01
[Security][tentative] Medium (correctness): Deposit Drain Rebuilds SSZ Subtree For Each Popped Deposit


evonide:
## Context

Summit stores accepted deposit requests in consensus state and processes only up to `max_deposits_per_epoch` at each penultimate epoch block. The same state also maintains an SSZ proof tree for the deposit queue, so queue mutations should keep finalizer work proportional to the configured cap and changed entries.

## Claim

A large accepted EL deposit backlog accumulates while `max_deposits_per_epoch` remains within the valid runtime cap. At each penultimate block, the finalizer pops up to the cap, and every pop rebuilds the deposit subtree over the still-queued deposits before the block can finish. Draining deposits pops from the front of a `VecDeque`, and every pop rebuilds the entire remaining deposit SSZ subtree. With a large backlog and a valid cap up to 256, one penultimate block can perform roughly cap times backlog SSZ rebuild work inside the finalizer path.

## Flow

The code path is reachable from accepted deposit requests already in `ConsensusState.deposit_queue`. The scan did not quantify the backlog size needed for material delay or prove malicious public ingress, so validation should measure finalizer occupancy under realistic deposit backlog sizes.

## Impact

Operators may expect the per-epoch deposit cap to bound deposit processing cost, but a large backlog can still keep the finalizer actor busy during penultimate-block execution. That can delay aux-data service, block execution, and epoch progress.

## Root Cause

The deposit proof tree represents the queue positionally, so front pops shift every remaining item and the implementation rebuilds the whole subtree for each individual pop instead of applying one batched rebuild after capped draining.

## Code

- Accepted deposit requests are appended to the deposit queue during execution-request parsing: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L1537.
- Penultimate-block processing pops up to `max_deposits_per_epoch` deposits: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/finalizer/src/actor.rs#L1728.
- Each `ConsensusState::pop_deposit` removes one item and immediately updates the SSZ tree: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/consensus_state.rs#L532.
[...]

PR #413, opened 2026-06-16
fix: batch deposit-queue SSZ rebuild on capped drain

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/412.

Addresses https://github.com/SeismicSystems/summit/issues/317.

Changes:
- types/consensus_state.rs: add pop_deposit_deferred() (pops the queue front, leaves the deposit subtree untouched) and rebuild_deposit_tree() (one rebuild_deposits over the current queue)
- finalizer/actor.rs: the penultimate-block drain loop now pops via pop_deposit_deferred(), tracks whether anything drained, and calls rebuild_deposit_tree() once after the loop; the subtree is stale only between the drain and the end-of-block root capture, which never reads it. Per-block rebuild is now O(B)
- tests: add test_deferred_deposit_drain_matches_per_pop (backlog 64, cap 16: deferred single-rebuild root == per-pop root == full rebuild) and test_deferred_deposit_drain_empties_small_backlog (backlog < cap, mirrors the break-on-empty)

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -2545,11 +2545,17 @@ async fn process_execution_requests<
     consts: &ProtocolConsts,
 ) {
     if is_penultimate_block_of_epoch(state.get_epocher(), new_height) {
+        // pop deposits without rebuilding the deposit subtree per pop, then
+        // rebuild once after the loop. front removal shifts every remaining
+        // item, so a per pop rebuild is o(cap * backlog) inside this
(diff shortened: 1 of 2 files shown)


sebastian-osec:
LGTM. This batches penultimate-block deposit draining so the deposit SSZ subtree is rebuilt once after the capped drain, removing the per-popped-deposit rebuild amplification from issue #317.

Issue #330, opened 2026-06-01
[Security][tentative] Medium (correctness): Genesis Validator Accounts Missing From Initial Proof Root


evonide:
## Context

`ConsensusState` keeps a live SSZ tree and a frozen proof snapshot. The frozen snapshot supplies `parent_beacon_block_root` for block aux data and backs the state-proof RPC, so it should describe the same consensus state that the node has installed.

Genesis startup builds the initial consensus state, then materializes the genesis validators as active validator accounts. That launch-time active set is the first committee and should be reflected in the root/proof state exposed before the first later capture.

## Claim

During normal startup without a checkpoint, Summit constructs an empty-validator `ConsensusState`, freezes/rebuilds the initial proof root, then inserts every genesis validator into live state; early block aux-data or proof requests can read the stale frozen snapshot before block execution captures a root containing those accounts. Startup initializes `state_root`, `proof_tree`, and `proof_validator_keys` before inserting genesis validator accounts, and `set_account` updates only the live SSZ tree. Until the first `capture_state_root`, the exposed frozen proof root can omit the genesis validator accounts that exist in live state.

## Flow

The path requires genesis startup rather than checkpoint startup, and a consumer of `get_state_root`, proposal aux data, or `getStateProof` before the first successful state-root capture after block execution. Static review did not prove a long-lived finalized divergence after the first capture.

## Impact

Early block or proof consumers can see a state root and proof tree that do not commit to the installed genesis validator accounts. This can make launch-time proof material misleading and can cause first-block `parent_beacon_block_root` semantics to differ from the live validator state used by consensus.

## Root Cause

Genesis validator insertion mutates only the live SSZ tree after `ConsensusState::new` has already rebuilt the frozen proof snapshot.

## Code

- `ConsensusState::new` initializes empty proof metadata and then calls `rebuild_ssz_tree`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L111.
- `rebuild_ssz_tree` updates the exposed `state_root` and `proof_tree` from current fields: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/consensus_state.rs#L753.
- Genesis startup adds validators only after `ConsensusState::new` returns: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/args.rs#L726.
[...]

PR #414, opened 2026-06-16
fix: commit genesis validators to the proof snapshot at startup

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/413 (which builds on https://github.com/SeismicSystems/summit/pull/412).

Addresses https://github.com/SeismicSystems/summit/issues/330.

Changes:
- node/args.rs: call state.rebuild_ssz_tree() after the genesis validator set_account loop so the frozen snapshot commits to the committee from the first block
- types/consensus_state.rs: rebuild_ssz_tree now also refreshes proof_validator_keys (alongside state_root and proof_tree) so the frozen proof snapshot is internally consistent everywhere rebuild runs; the decode-with-capture override in read_cfg still runs after rebuild, so capture-time keys win on restart (no regression)
- tests: add test_genesis_materialization_refreshes_proof_snapshot (frozen root is stale before the rebuild, then state_root == proof_tree().root() == ssz_tree().root() and proof_validator_keys == the sorted committee after)

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -976,6 +976,12 @@ fn get_initial_state(
             };
             state.set_account(pubkey_bytes, account);
         }
+        // ConsensusState::new froze the proof snapshot over an empty validator
+        // set before these genesis accounts were inserted, and set_account only
+        // touches the live tree. Re-freeze so get_state_root / proof_tree commit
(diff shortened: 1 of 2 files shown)


sebastian-osec:
This looks good to me.

Issue #338, opened 2026-06-01
[Security][tentative] Medium (attackable): Orchestrator Awaits Catch-Up Hints Inline


evonide:
## Context

Future-epoch pending traffic is only advisory catch-up input. Handling it should not block the same loop that must process epoch transitions.

## Claim

An authenticated peer that can send future-epoch pending backup traffic can make the orchestrator await syncer catch-up hint delivery inline while epoch transition messages wait, if the syncer mailbox is already slow or full. The evidence proves the peer-controlled entry into the awaited hint path, but not that one peer can independently create the required syncer backpressure. The orchestrator violates the epoch-control invariant by awaiting `syncer.hint_finalized(...)` inline while handling future-epoch pending-mux backup traffic. A slow or full syncer mailbox can therefore park the orchestrator before it processes `Enter` or `Exit` messages needed to register schemes and run the next epoch.

## Flow

This is reachable on the orchestrator pending-mux backup path when a peer advertises a later epoch and the syncer mailbox send is slow or blocked, for example because the bounded mailbox is saturated by other syncer work.

## Impact

Future-epoch pending traffic combined with syncer mailbox saturation or slowness can stall the orchestrator inside the catch-up hint path. While stalled, the orchestrator does not process epoch `Enter`/`Exit`, so a finalized epoch transition may not register schemes or start the next Simplex engine. This can halt liveness at epoch boundaries for the affected node.

## Root Cause

The implementation misses the invariant check or state update described in the claim, allowing the accepted state to be consumed later as if it were valid.

## Code

- Future-epoch pending backup messages are handled inline and call `syncer.hint_finalized(...).await`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/orchestrator/src/actor.rs#L198, https://github.com/SeismicSystems/summit/blob/ed2c5c8/orchestrator/src/actor.rs#L218.
- The same select loop processes `Enter`/`Exit` in a later arm: https://github.com/SeismicSystems/summit/blob/ed2c5c8/orchestrator/src/actor.rs#L221, https://github.com/SeismicSystems/summit/blob/ed2c5c8/orchestrator/src/actor.rs#L228, https://github.com/SeismicSystems/summit/blob/ed2c5c8/orchestrator/src/actor.rs#L258.
[...]

PR #415, opened 2026-06-17
fix: don't block the orchestrator loop on advisory catch-up hints

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/338.

Changes:
- syncer/ingress/mailbox.rs: add try_hint_finalized(height, targets), a non-blocking try_send variant; drop the advisory hint with a warn on a full mailbox (the ahead peer re-advertises and the finalization also arrives via the normal flow), error-log on a closed receiver
- orchestrator/actor.rs: pending_backup arm now calls try_hint_finalized(...) (no .await) so a full syncer mailbox can't park the loop ahead of Enter/Exit
- tests: add try_hint_finalized_drops_when_mailbox_full (capacity-1 mailbox; second hint returns synchronously and is dropped, only the first is enqueued)

### DIFF
--- a/orchestrator/src/actor.rs
+++ b/orchestrator/src/actor.rs
@@ -216,7 +216,13 @@ where
                 // multiple peers claiming to be ahead, each call adds them to the target set,
                 // giving us more peers to try fetching from.
                 let boundary_height = self.epocher.last(our_epoch).expect("epoch should exist");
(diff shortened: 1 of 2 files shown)


sebastian-osec:
LGTM. This removes the blocking await from the advisory catch-up hint path, so syncer mailbox backpressure no longer stalls orchestrator epoch transitions.

Issue #318, opened 2026-06-01
[Security][tentative] Medium (correctness): E2E Scenarios Can Pass While Required Participants Are Failed Or Unchecked


evonide:
## Context

Because these binaries are used to support findings, their success predicates must prove the required participant set stayed healthy. A helper that skips an affected validator or ignores a node task failure can turn a broken scenario into misleading validation evidence.

## Claim

A maintainer, CI job, or validation workflow runs one of the affected E2E scenarios while the skipped source, withdrawn, observer, or proof participant has crashed, stalled, or failed to catch up. The scenario's remaining RPC assertions can still pass, and the wrapper can then call `std::process::exit(0)` after logging ignored node-handle or thread failures. Several E2E scenario binaries can report success without checking all required participants: checkpoint scenarios intentionally stop node0 and then exclude it from final progress checks, withdrawal scenarios omit the withdrawn node from progress checks and verify removal only through node0, and the wrappers print node task completion or thread join failures without converting them into a nonzero exit.

## Flow

Reachable through the configured validation scenarios because they report scenario success based on selected end conditions without consistently proving that every required participant stayed healthy. A failed or omitted participant can therefore be hidden by a scenario that still reaches its narrow success marker.

## Impact

The main impact is integrity of validation, not live consensus. A finding can be promoted or considered exercised by a stock scenario that did not actually prove the affected participant remained healthy, caught up, exited, or observed the expected state. That can mask regressions and produce false confidence in checkpoint, withdrawal, observer, and proof workflows.

## Root Cause

The scenario wrappers do not bind scenario success to all required node outcomes. They use partial progress loops, accept node-handle completion as a log event, ignore thread join errors, and force process success after local assertions.

## Code

- `stake-and-checkpoint` stops node0 for checkpoint copying, then checks only nodes `1..num_nodes` before printing success: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/bin/stake_and_checkpoint.rs#L359, https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/bin/stake_and_checkpoint.rs#L576, https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/node/src/bin/stake_and_checkpoint.rs#L598.
[...]

PR #416, opened 2026-06-17
fix: fail e2e scenarios when a required node dies

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/318.

Changes:
- node/args.rs: run_node_local now returns `Handle<anyhow::Result<()>>` and run_node_local_inner returns the supervise_node_tasks outcome instead of logging-and-swallowing it. Still no `process::exit` there (would orphan the child reth processes) — the caller decides.
- e2e scenario binaries (`stake_and_checkpoint`, `stake_and_join_with_outdated_ckpt`, `withdraw_and_exit`, `sync_from_genesis`, `observer`, `verify_consensus_state_proof`, `protocol_params`): each node spawn closure now fails the scenario (`eprintln` + `exit(1)`) when run_node_local returns `Err`. Safe against the intentionally-exiting withdrawn/stopped nodes, which take a coordinated-shutdown path and return `Ok`. Required making the spawn closures `move` (the message captures the Copy node index).
- node/bin/testnet.rs: `try_join_all` now inspects each node's inner Result and logs failures.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -593,7 +593,7 @@ pub fn run_node_local(
     flags: RunFlags,
     checkpoint: Option<ConsensusState>,
     checkpoint_parent_block: Option<Block>,
-) -> Handle<()> {
+) -> Handle<anyhow::Result<()>> {
     context.spawn(async move |context| {
         let key_store = expect_key_store(&flags.key_store_path);
         run_node_local_inner(
@@ -603,7 +603,7 @@ pub fn run_node_local(
             checkpoint,
             checkpoint_parent_block,
         )
-        .await;
+        .await
     })
 }
 
@@ -613,7 +613,7 @@ async fn run_node_local_inner(
     key_store: KeyStore<PrivateKey>,
     checkpoint: Option<ConsensusState>,
     checkpoint_parent_block: Option<Block>,
-) {
+) -> anyhow::Result<()> {
     let context = context.with_label("summit_cw");
 
     let genesis = acquire_genesis(&context, &flags).await;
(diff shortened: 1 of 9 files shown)


sebastian-osec:
I think it is not completely fixed yet.

Some of the predicates are still partial. For example, `stake_and_checkpoint` and `stake_and_join_with_outdated_ckpt` still skip node0 in the final progress loop, and `withdraw_and_exit` / `sync_from_genesis` still check progress only for `0..(NUM_NODES - 1)`. That may be intentional for stopped/withdrawn nodes, but the scenario still does not explicitly assert the expected shutdown/removal outcome for every omitted participant.

There are also still join paths that only print thread join failures and then continue to “All nodes shut down cleanly” / `exit(0)`, for example in `observer` and the checkpoint scenarios. `testnet.rs` similarly logs inner node errors from `try_join_all` but does not fail the process.

Each scenario should define the required participant set and either assert progress for each required node or assert the expected clean shutdown/removal for intentionally stopped/withdrawn nodes, and join failures should become nonzero scenario failures.

matthias-wright:
Update(b4507b1):
- Fail e2e scenarios on unexpected node-handle exit or join failure

sebastian-osec:
Thanks.

I think there is one remaining similar path in `protocol_params.rs`: the node wrapper still treats `_ = node_handle.fuse()` as just `println!("Node {} handle completed", x)`. Since there is no expected withdrawn/stopped participant in that scenario, an early handle completion should probably be treated the same way as the other scenarios and fail the scenario.

So I think the update fixes my comment for the touched paths, but I would still patch `protocol_params.rs` before closing #318 completely.

matthias-wright:
Update(018839f):
- Fail e2e scenarios on remaining unguarded node-exit paths

sebastian-osec:
This looks good to me.

Issue #314, opened 2026-06-01
[Security][tentative] Medium (attackable): Consensus State Nested Counts Allocate Before Byte Bounds


evonide:
## Context

Summit decodes `ConsensusState` from checkpoints and finalizer DB values during startup and recovery. Decode paths should bound collection allocation by the number of complete elements that can be present, not by attacker-controlled element counts.

## Claim

Several nested `ConsensusState` and related decoders use `len.min(buf.remaining())` as a collection capacity cap. `buf.remaining()` is a byte count, not an element count, so a malformed state can still allocate far more element slots than can be validly decoded before later returning `EndOfBuffer`.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

Startup or recovery can suffer memory pressure or process failure while parsing malformed checkpoint or persisted state bytes that should be rejected cheaply. The severity is limited by the checkpoint/local persistence trust boundary unless another remotely reachable state-byte ingress is found.

## Root Cause

Decode code treats remaining bytes as if they were a safe upper bound on element count. For multi-byte elements, a byte bound can still permit many impossible elements and allocate before checking each element's full encoded size.

## Code

- `deposit_queue` and `protocol_param_changes` allocate with `len.min(buf.remaining())`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/consensus_state.rs#L844-L858.
- Added-validator inner vectors, removed validators, and pending execution requests repeat the same pattern: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/consensus_state.rs#L878-L909.
- Withdrawal schedule pubkey queues use a byte count as element capacity: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/withdrawal.rs#L388-L400.
- `DynamicEpocher` segments use the same allocation pattern: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/dynamic_epocher.rs#L212-L220.
[...]

PR #417, opened 2026-06-17
fix: bound decode allocation by decoded elements, not remaining bytes

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/314.

Changes:
- types/consensus_state.rs — drop the capacity hint for deposit_queue, protocol_param_changes, added_validators inner, removed_validators, pending_execution_requests
- types/withdrawal.rs — drop it for the schedule pubkeys queue
- types/dynamic_epocher.rs — drop it for segments (kept the existing segments_len == 0 reject)
- Collections now grow on push; the loops already bail on the first EndOfBuffer, so only genuinely decoded elements are ever allocated
- Added 3 regression tests asserting a huge nested count fails cheaply without pre-allocating (consensus_state, withdrawal schedule, dynamic_epocher)

### DIFF
--- a/types/src/consensus_state.rs
+++ b/types/src/consensus_state.rs
@@ -1248,7 +1248,14 @@ impl Read for ConsensusState {
         let latest_height = buf.try_get_u64().map_err(|_| Error::EndOfBuffer)?;
 
         let deposit_queue_len = buf.try_get_u32().map_err(|_| Error::EndOfBuffer)? as usize;
-        let mut deposit_queue = VecDeque::with_capacity(deposit_queue_len.min(buf.remaining()));
+        // Never pre-size collections from the decoded length prefixes below:
+        // they are attacker-controlled u32s, and `buf.remaining()` is a byte
+        // count rather than an element count, so even a `min(buf.remaining())`
+        // hint over-allocates by `size_of::<T>()` per slot before any element
+        // is validated. Growing on push is safe — each loop reads from `buf`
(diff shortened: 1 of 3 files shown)


sebastian-osec:
This looks good to me.

One small remaining allocation-hardening case in the same decoder: `captured_bytes` still allocates from a decoded length before checking that the buffer actually contains that many bytes:

```rust
let len = buf.try_get_u32().map_err(|_| Error::EndOfBuffer)? as usize;
let mut bytes = vec![0u8; len];
buf.try_copy_to_slice(&mut bytes)
    .map_err(|_| Error::EndOfBuffer)?;
```

A malformed checkpoint/state blob could set `has_captured = true` and a huge `len`, causing allocation before the later `EndOfBuffer`. Could we apply the same guard used for `pending_execution_requests` here?

```rust
if len > buf.remaining() {
    return Err(Error::EndOfBuffer);
}
```

matthias-wright:
Update(210a427):
- Bound captured_bytes decode allocation by remaining bytes

sebastian-osec:
Looks good to me.

I re-checked the broader #314 pattern, and the risky `with_capacity(len.min(buf.remaining()))` sites appear to be removed from the affected decoders. The remaining length-prefixed byte allocations are guarded before allocation.

Optional test nit: a few of the regression tests stop immediately after the huge count, so `buf.remaining() == 0`; the old code would also have allocated zero in those exact cases. Adding some trailing bytes would make the tests reproduce the original over-allocation shape more directly.

Issue #220, opened 2026-05-22
[Security][tentative] High (attackable): Generated Key Files Use Default Permissions


evonide:
## Context

Summit key generation creates the local node identity key and the BLS consensus signing key that validators later load from the keystore path. Those private-key files are a local trust boundary: only the node operator account should be able to read them.

On Unix-like systems, file creation through default filesystem APIs is mediated by the process umask. This finding concerns generated keystores where Summit relies on those defaults instead of explicitly creating private-key material with owner-only permissions.

## Claim

`summit keys generate` writes private keys with default filesystem modes, so a permissive Unix umask can create group- or world-readable validator key files.

A malicious local user or co-tenant process with access to the keystore path can read newly generated `node_key.pem` or `consensus_key.pem` when Summit inherits a permissive operator umask, turning default file modes into node-identity or validator-signing-key exposure.

## Flow

The path is reachable on Unix-like systems when an operator runs `summit keys generate` with a permissive umask such as `022` and the resulting keystore path is accessible to another local principal. It is not a remote read primitive by itself; the attacker must already have local filesystem reachability to the generated files or their containing directory.

## Impact

Stealing `consensus_key.pem` allows unauthorized consensus signing; stealing `node_key.pem` compromises node identity. This can cause equivocation, unauthorized participation, or targeted validator disruption.

## Root Cause

Private-key material is written through default filesystem APIs instead of creating directories and files with explicit owner-only permissions.

## Code

- Key generation creates the keystore directory with `fs::create_dir_all`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/keys.rs#L98.
- It generates the Ed25519 node private key: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/keys.rs#L101.
- It writes `node_key.pem` with `fs::write`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/keys.rs#L105.
- It generates the BLS consensus private key: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/keys.rs#L107.
- It writes `consensus_key.pem` with `fs::write`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/keys.rs#L111.
- Runtime loading reads those same private-key files: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/key_paths.rs#L68, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/key_paths.rs#L78.
[...]

PR #418, opened 2026-06-17
fix: harden keystore file permissions for generated keys

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/220.

Changes:
- node/src/keys.rs
  - create_keystore_dir() — create the keystore dir 0700 on Unix (DirBuilder.mode), create_dir_all fallback elsewhere
  - write_private_key_file() — create key files 0600 via OpenOptions.mode, plus an explicit set_permissions(0600) so the overwrite path also re-tightens a pre-existing loose file (truncate preserves the old mode); fs::write fallback on non-Unix
- types/src/key_paths.rs — both key readers now call warn_if_permissions_too_open(): Unix-only tracing::warn! when mode & 0o077 != 0. Warns rather than rejects so a validator with legacy modes isn't taken offline on restart (chmod 600 clears it).
- Tests (#[cfg(all(test, unix))]) — assert generated dir + both key files are owner-only and still load; assert overwrite tightens a seeded 0644 file back to 0600. No new dependency.

### DIFF
--- a/node/src/keys.rs
+++ b/node/src/keys.rs
@@ -2,6 +2,7 @@ use anyhow::Result;
 use clap::{Args, Subcommand};
 use std::fs;
 use std::io::{self, Write};
+use std::path::Path;
 
 use commonware_codec::Encode;
 use commonware_cryptography::Signer;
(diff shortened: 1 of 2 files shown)


sebastian-osec:
LGTM.

Issue #235, opened 2026-05-22
[Security][tentative] Low (attackable): Block Codec Allows Unbounded Remote Decode Allocation


evonide:
## Context

The syncer accepts encoded blocks from peers during catch-up and backfill, then decodes them before higher-level consensus and application checks decide whether they are valid. Codec ingress is therefore a resource boundary: peer-supplied length fields should be bounded before the node allocates buffers or spends decode work on unauthenticated data.

## Claim

Summit's block `Read` implementation trusts a length prefix up to the remaining input and allocates the declared payload before protocol validation. Inner variable fields also lack Summit-specific caps, so oversized peer-supplied blocks can consume memory and decode CPU before rejection. The same allocation pattern exists in `Header::read`, but the remote sync/backfill path decodes blocks or certificate/block tuples rather than standalone headers through that codec.

A malicious sync or backfill peer can answer a block, finalized, or notarized request with a length-prefixed block payload that fits under the configured P2P message limit but is oversized for Summit's protocol object limits; the receiver allocates and decodes the full block body before semantic validation rejects it, creating remote memory and CPU DoS pressure.

## Flow

The block path is reachable on remote syncer response handling: `Request::Block` decodes `B` directly, and `Request::Finalized`/`Request::Notarized` decode a certificate plus `B`, where Summit instantiates `B` as `Block`. The peer must be able to supply a length-prefixed block payload whose length fits in the received buffer and under the configured P2P `max_message_size_bytes`, but exceeds Summit's intended protocol object sizes. Standalone `Header::read` is allocation-heavy under direct codec use, but the supported remote delivery path is the block/certificate-block decoder.

## Impact

A peer can send oversized block payloads that allocate large buffers and spend CPU decoding before later validation rejects them. This creates a remote DoS surface against sync/backfill and can amplify existing buffering risks. The P2P path is message-bound rather than truly unbounded: it can repeatedly force up-to-limit allocation work, but one peer response is still constrained by the configured message-size cap.

## Root Cause

Codec ingress uses generic length-prefixed reads and SSZ variable data decoding without Summit-specific maximum encoded sizes for blocks, execution requests, or header fields embedded inside blocks.

## Code

- `Block::read` uses the same length-prefixed allocation pattern: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L265, https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/block.rs#L270.
[...]

PR #419, opened 2026-06-17
chore: avoid redundant payload copy in Block::read

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/235.

Changes:
- types/src/block.rs — Block::read now decodes SSZ directly from the buffer's contiguous chunk (buf.chunk()[..len] + buf.advance(len)) on the hot path (&[u8]/Bytes), instead of first copying the up-to-message-size payload into a temporary Vec. A contiguous-copy fallback is kept for non-contiguous buffers. The existing len > buf.remaining() guard is unchanged.

Note:
This PR doesn't add any max encoded block/header size enforcement, because we already have `max_message_size_bytes`. Furthermore, the `len > buf.remaining()` check rejects an oversized prefix before any allocation.

### DIFF
--- a/types/src/block.rs
+++ b/types/src/block.rs
@@ -249,16 +249,29 @@ impl Read for Block {
     type Cfg = ();
 
     fn read_cfg(buf: &mut impl Buf, _cfg: &Self::Cfg) -> Result<Self, Error> {
-        let len: u32 = buf.try_get_u32().map_err(|_| Error::EndOfBuffer)?;
-        if len as usize > buf.remaining() {
+        let len = buf.try_get_u32().map_err(|_| Error::EndOfBuffer)? as usize;
+        if len > buf.remaining() {
             return Err(Error::Invalid("Block", "improper encoded length"));
         }
 
-        let mut payload = vec![0u8; len as usize];
-        buf.try_copy_to_slice(&mut payload)
-            .map_err(|_| Error::EndOfBuffer)?;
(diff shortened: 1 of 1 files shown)


sebastian-osec:
Thanks, that clarification makes sense.

I agree this PR improves `Block::read`: on the contiguous-buffer path it avoids the extra full-payload `Vec` allocation/copy before SSZ decoding. I also agree that `max_message_size_bytes` bounds the raw inbound message size, and `len > buf.remaining()` prevents allocating beyond the received buffer length.

My remaining question is about the intended boundary.

If we’re treating `max_message_size_bytes` as the accepted upper bound for encoded blocks, then this looks fine to me as a performance improvement.

If we are treating `max_message_size_bytes` as a stricter codec-level limits, then I think this is still a partial fix. A peer can send a block payload close to `max_message_size_bytes`; `Block::read` will avoid the extra outer copy, but SSZ decoding can still allocate for inner fields like execution requests, `extra_data`, added validators, and removed validators. That means peak memory and CPU work during decode can still be higher than the raw message size.

matthias-wright:
I agree that the SSZ decoding can create a memory footprint that exceeds `max_message_size_bytes`.
But since the size of a decoded SSZ object is bounded linearly by the encoded size, I'm not worried about the overhead.
I think we can rely on `max_message_size_bytes` to do the heavy lifting here. We just have to set it to a reasonable size.

Issue #233, opened 2026-05-22
[Security][tentative] Medium (attackable): Withdrawal Field Proofs Are Not Bound To The Requested Pubkey


evonide:
## Context

The state-proof API offers withdrawal-field proofs selected by validator pubkey so external consumers can verify individual pending-withdrawal fields against a consensus-state root. For that API to be trustless, the proof statement must bind the requested pubkey to the returned field value.

The normal flow is that Summit uses the pubkey to locate the pending withdrawal, emits a Merkle proof, and a downstream verifier checks that proof without trusting the proof provider. This finding concerns the missing selector binding in the proof object that verifier receives.

## Claim

`getStateProof` supports `withdrawal_field:<pubkey>:<field>` requests, but the returned proof authenticates only a positional field leaf. It does not prove that the field belongs to the requested withdrawal pubkey, and the response neither binds nor echoes the request descriptor.

A malicious proof provider or intermediary can answer a by-pubkey withdrawal-field request with a valid positional proof for the same field from a different pending withdrawal under the same state root, causing a verifier to authenticate the field value without proving it belongs to the requested pubkey.

## Flow

The path is reachable for trustless consumers that request single-field withdrawal proofs and verify only the returned Merkle branch against the root. It requires at least two pending withdrawals where substituting one field value for another matters to the consumer, and it affects the external proof-consumer boundary rather than Summit's own consensus-state transition.

## Impact

External contracts, indexers, or accounting systems relying on single-field withdrawal proofs can be misled about payout address, amount, deduction, or epoch. The Merkle proof verifies, but it proves only that a positional withdrawal field exists, not that the field belongs to the requested pubkey.

## Root Cause

The by-pubkey selector is used only server-side to locate a position. The emitted proof and verifier API bind `gindex`, leaf, and branch to the root, but omit the pubkey field or canonical request path from the authenticated statement.

## Code

- Docs advertise `withdrawal_field:<pubkey>:amount` as a single-field proof for that pubkey: https://github.com/SeismicSystems/summit/blob/ed2c5c8/docs/ssz-merklization.md#L439.
- The parser stores the selector as `WithdrawalField(pubkey, field_index)`: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/ssz_tree_key.rs#L107.
- Finalizer generates only the requested field proof by looking up the pubkey locally: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L1132.
[...]

PR #420, opened 2026-06-17
fix: bind by-pubkey field proofs to the requested pubkey

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/233.

Changes:
- **types/ssz_state_tree.rs** — add `KeyedFieldProof` + `verify(root, key, fields_per_item)`; add `generate_withdrawal_keyed_field_proof_by_key` and `generate_validator_keyed_field_proof`; add `StateProofEntry { field, key: Option<_> }`.
- **types/consensus_state_query.rs, finalizer/{actor,ingress}.rs** — thread `StateProofEntry` through the `GenerateStateProof` response; by-pubkey field requests emit the keyed pair, all others `key: None`.
- **types/rpc.rs, rpc/server.rs** — add `key_proof: Option<SszProof>` to `StateProofResult` (`skip_serializing_if` → JSON back-compatible).
- **node/src/bin/verify_consensus_state_proof.rs** — TEST E: require `key_proof`, verify it on-chain, run the binding check, and assert a substituted pubkey is rejected.
- **docs/ssz-merklization.md** — document the binding requirement and `key_proof`.
- **tests** — new `withdrawal_keyed_field_proof_binds_to_requested_pubkey` covering honest binding, wrong-pubkey rejection, and the substitution forgery.

### DIFF
--- a/docs/ssz-merklization.md
+++ b/docs/ssz-merklization.md
@@ -405,6 +405,21 @@ The state root appears on-chain in EL block `el_block_number + 1`.
 
 Takes a list of key strings and returns the state root, EL block number, and one result for each key. Each result echoes the requested key and contains either an `SszProof` or an error for a key that is absent or out of range.
 
+> **Binding for by-pubkey field proofs.** A bare `SszProof` for a single field
+> (`validator_field:`/`withdrawal_field:`) proves only that *some* positional
+> leaf exists under the root — it does **not** prove the field belongs to the
+> requested pubkey, since the pubkey→position mapping happens server-side. A
+> malicious provider or intermediary could answer a by-pubkey request with the
+> same field from a *different* item under the same root. For these requests the
+> result therefore also carries a `key_proof`: a companion `SszProof` of the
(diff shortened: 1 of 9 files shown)


sebastian-osec:
This looks good to me for the main #233 issue.

The important invariant is now enforceable by the consumer: verify the field proof, verify the key proof, require the key leaf to equal the requested pubkey, and require both gindices to resolve to the same item. The added withdrawal test covers the real substitution case well: the bare positional proof from another withdrawal still verifies, but the keyed proof rejects it.

Optional hardening nit: `KeyedFieldProof::verify` checks that the key proof verifies, that its leaf equals the requested pubkey, and that it belongs to the same item as the field proof. It may also be worth checking that the key proof’s gindex is specifically the canonical pubkey field selector (`WITHDRAWAL_FIELD_PUBKEY` / `VALIDATOR_FIELD_NODE_PUBKEY`).

matthias-wright:
Update(80591b2):
- Require keyed field proof to address the canonical pubkey field

sebastian-osec:
thanks, looks good to me now.

Issue #305, opened 2026-06-01
[Security][tentative] Medium (correctness): Application Consensus Responses Wait Behind Syncer Mailbox Sends


evonide:
## Context

The application actor produces consensus-facing responses for local proposals and verification requests, while the syncer maintains block availability, repair, cache, archive, and broadcast state. These components communicate through bounded mailboxes, so syncer work can apply backpressure when its actor or downstream tasks fall behind.

Consensus timing depends on the application returning an already-built proposal digest or verification result promptly. Syncer notification is auxiliary propagation work and should not sit on the critical path after the application has enough information to answer Commonware.

## Claim

During local proposal, verification, or consensus activity, the bounded application/consensus-to-syncer mailbox can fill or the syncer actor can be delayed by inline cache, archive, broadcast, resolver, or dispatch work; if that happens after the application has already built or validated a block, `syncer.proposed` or `syncer.verified` can remain pending before the consensus proposal digest or vote response is sent, causing the validator to miss proposal or voting deadlines. The application actor awaits syncer mailbox sends before answering Commonware proposal and verification requests. This puts auxiliary syncer notification on the consensus response path, so a full or stalled syncer mailbox can delay an already-ready proposal digest or verification result.

## Flow

The path is reachable during normal proposal and verification handling. It requires the syncer mailbox send to block or the syncer actor to be delayed while consensus is waiting for the application response. The earlier peer-driven wording was too strong: network resolver traffic is received on `resolver_rx`, while the bounded mailbox awaited by `syncer.proposed` and `syncer.verified` carries internal application, consensus-engine, and control messages.

## Impact

A validator can miss proposal or vote deadlines despite already having the proposal digest or validation result. Under recovery or peer-pressure conditions, enough validators stalling on syncer enqueue can delay notarization or finality.

## Root Cause

Consensus responses are ordered after best-effort syncer notification instead of being returned as soon as application-local proposal or verification work completes.

## Code

- Proposal handling sends the block to syncer before returning the digest: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L167, https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L171.
[...]

PR #421, opened 2026-06-18
fix: keep syncer notification off the consensus response path

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/305.

Changes:
- **application/src/actor.rs (propose):** return the digest to consensus first,
  then dispatch `syncer.proposed(...)` in a spawned `"proposed"` task. Delivery is
  still guaranteed (not lossy), so the block is reliably cached and broadcast, but
  it no longer sits on the proposal-response path or blocks the application loop.
- **application/src/actor.rs (verify):** send the vote before `syncer.verified(...).await`
  (already runs in the spawned verify task), so the bounded send no longer gates
  the vote response.

### DIFF
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -161,11 +161,22 @@ impl<
                                                     "proposed block"
                                                 );
 
-                                                // send block to syncer for caching and broadcasting
-                                                syncer.proposed(round, block).await;
-
-                                                // send digest to consensus
+                                                // Return the digest to consensus immediately. The
+                                                // block is already built, so caching and broadcasting
+                                                // via the syncer is auxiliary work that must not sit on
+                                                // the proposal response path or block the application
+                                                // loop behind a full or slow syncer mailbox.
+                                                // It runs off the loop with guaranteed (not lossy)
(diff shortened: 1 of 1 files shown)


sebastian-osec:
Mostly LGTM. This fixes the direct #305 propose/verify response-path issue: the application now sends the proposal digest / vote response before awaiting `syncer.proposed(...)` or `syncer.verified(...)`.

One small remaining concern: `Plan::Forward` still awaits `syncer.forward(...)` on the main application actor loop. Since `syncer.forward` also sends into the bounded syncer mailbox, a full or stalled syncer could still block the application actor while handling `Broadcast(Forward)`, delaying later `Propose` or `Verify` messages.

So I think the direct fix is good, but it would be worth either moving `Plan::Forward` off-loop too, or adding a regression test showing a full syncer mailbox cannot delay subsequent propose/verify responses.

matthias-wright:
Update(d552b59):
- Keep all syncer enqueues off the application loop

sebastian-osec:
LGTM now

Issue #214, opened 2026-05-22
[Security][tentative] High (attackable): Checkpoint Startup Accepts Unauthenticated Artifacts


evonide:
## Context

Checkpoint startup lets a fresh or joining node initialize consensus, syncer, and finalizer state from checkpoint artifacts instead of replaying from genesis. The trust boundary is the artifact supplier: startup is expected to verify that imported checkpoint data is bound to finalized history before using it as authoritative recovery state.

## Claim

Checkpoint startup can install checkpoint artifacts that are not fully authenticated by finalized history. Standalone checkpoint imports convert a single file into `ConsensusState` while skipping `verify_checkpoint_chain`, and directory imports pass optional `last_block` and top-level `finalized_header` artifacts to startup without binding them to the verified checkpoint/header chain.

A selected checkpoint artifact supplier can give a fresh or checkpoint-joining node a standalone checkpoint file without a finalized-header chain, or a directory checkpoint with unbound optional artifacts; startup then skips or narrows verification and seeds consensus, syncer, or finalizer state from unauthenticated or mismatched recovery data. This is a local startup/import path, not a public RPC or P2P write path into an already-running node.

## Flow

This is reachable on checkpoint startup with a fresh or checkpoint-joining database when the operator supplies `--checkpoint-path`. Single-file imports require no finalized-header chain; directory imports require control over the checkpoint directory contents. No source evidence shows that a public RPC caller, P2P peer, or passive checkpoint mirror can force an already-running node to consume these artifacts.

## Impact

A malicious or malformed standalone checkpoint file can provide a self-consistent but unsigned state and be accepted as the node's initial consensus state on a fresh database.

Separately, a malicious checkpoint directory can pair a verified checkpoint/header chain with unrelated optional `last_block` or top-level finalized-header artifacts, seeding sync/finalizer state with data not bound to the verified checkpoint target.

This weakens checkpoint startup from "verified finalized history" to "locally well-formed artifacts plus optional verification of only some files."

## Root Cause

For the standalone checkpoint path, startup verification is conditional on both checkpoint and header-chain artifacts being present, while single-file checkpoint startup can never supply a header chain through that file path.

For the source-backed directory path, import verification only covers the raw checkpoint plus `finalized_headers_chain`; it does not bind every loaded artifact to the verified chain target.

## Code

[...]

PR #422, opened 2026-06-18
fix: require checkpoint verification at startup and bind import artifacts to verified chain

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/402.

Addresses https://github.com/SeismicSystems/summit/issues/214.

Changes:
- **Require verification by default.** `run_node_inner` now panics when a checkpoint is supplied without a finalized-headers chain to verify against, instead of warning and continuing.
- **Add `--unsafe-skip-checkpoint-verification`** (`RunFlags`, requires `--checkpoint-path`) as the explicit opt-out for importing a checkpoint without verification artifacts; logs a loud UNSAFE warning when used.
- **Bind `last_block` to the verified chain.** On the verified path, require `last_block.digest() == terminal.finalization().proposal.payload` (the verified chain terminal), panicking on mismatch.
- **Use the signature-verified terminal as the completion finalization.** Replace the separately-loaded, unverified top-level `finalized_header` with the chain terminal `FinalizedHeader` (whose certificate was verified by `verify_checkpoint_chain`) before handing it to the syncer.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -181,6 +181,17 @@ pub struct RunFlags {
     )]
     pub weak_subjectivity_header_digest: Option<String>,
 
+    /// Import a checkpoint WITHOUT verifying it against a finalized-header chain.
+    ///
+    /// UNSAFE: the imported consensus state is trusted entirely from the supplied
+    /// checkpoint artifact. By default, checkpoint startup requires a checkpoint
+    /// directory containing finalized_headers/ together with
+    /// --weak-subjectivity-epoch and --weak-subjectivity-header-digest. Set this
+    /// flag to bypass that requirement (e.g. to import a standalone checkpoint
+    /// file). Only use it when the checkpoint source is fully trusted.
+    #[arg(long, requires = "checkpoint_path")]
+    pub unsafe_skip_checkpoint_verification: bool,
+
     /// IP address for this node (optional, will use genesis if not provided)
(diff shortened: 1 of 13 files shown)


sebastian-osec:
Mostly LGTM. The main #214 fix looks right to me: checkpoint startup now requires verification by default, binds `last_block` to the verified chain terminal, and uses the verified terminal finalized header instead of trusting the separately loaded top-level `finalized_header`.

A couple of nits / follow-ups:

- I’d add direct regression tests for the startup cases: standalone checkpoint rejected without `finalized_headers/`, accepted only with `--unsafe-skip-checkpoint-verification`, directory checkpoint rejected when `last_block` does not match the verified terminal finalization payload, and mismatched top-level `finalized_header` being ignored/replaced by the verified terminal header.
- The invariant is enforced in `run_node_inner`, but `EngineConfig` / `SyncCheckpoint` and the syncer still accept `checkpoint_last_block` and `checkpoint_finalized_header` independently. That does not look like a bypass for the current CLI checkpoint startup path, but enforcing or asserting the same binding closer to `SyncCheckpoint` / `store_finalization` would make the boundary harder to misuse later.

matthias-wright:
Update(416b5bd):
- Cover checkpoint startup policy and harden artifact binding

sebastian-osec:
Looks good to me after the latest commits.

Issue #216, opened 2026-05-22
[Security][tentative] High (attackable): Checkpoint Transition Queues Are Not Bound To The Signed Terminal Header


evonide:
## Context

Summit tracks validator-set changes through finalized-header deltas and pending transition queues in consensus state. A checkpoint importer should only install pending add/remove queues that are consistent with the signed terminal header and the finalized history used to reconstruct the active participant set.

## Claim

Checkpoint verification applies finalized-header validator deltas to a temporary participant set, but does not bind the decoded checkpoint state's pending `added_validators` and `removed_validators` queues to the transition history authenticated by the signed finalized headers.

A malicious checkpoint provider that can provide an otherwise valid finalized-header chain whose terminal header already signs the supplied checkpoint blob can include pending validator-transition queues that are not bound to the signed terminal header deltas. A passive public distributor cannot add or change these queues in an honest checkpoint without changing `sha256(checkpoint.data)` and failing terminal-header verification.

## Flow

The verified-chain path requires checkpoint import/verification, a finalized-header sequence accepted by the checkpoint verifier, a terminal header that authenticates the supplied checkpoint bytes, and decoded checkpoint state whose active membership still passes the final membership check while pending add/remove queues diverge from the header-authenticated transition history.

## Impact

A checkpoint can verify against the finalized header chain while carrying pending add/remove queues that produce a different next committee after import. The node can then diverge from the validator set reconstructed from finalized headers, undermining checkpoint soundness and future certificate verification.

## Root Cause

Checkpoint verification authenticates active membership through the finalized headers but omits an equality check between serialized checkpoint transition queues and the terminal header's committed validator deltas.

## Code

- Last-block aux data puts next-epoch `added_validators` and current removals into the Summit block header: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L962.
- Application verification checks a block header's validator deltas against local aux data: https://github.com/SeismicSystems/summit/blob/ed2c5c8/application/src/actor.rs#L670.
- Checkpoint verifier applies header deltas only to an accumulated temporary participant set: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L249.
[...]

PR #423, opened 2026-06-18
fix: bind checkpoint transition queues to the verified terminal header

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/422 (which builds on https://github.com/SeismicSystems/summit/pull/402)

Addresses https://github.com/SeismicSystems/summit/issues/216.

Changes:
- **types/src/checkpoint.rs**: add Step 5 to `verify_checkpoint_chain_with_weak_subjectivity` binding the decoded queues to the terminal header's committed deltas — require `checkpoint_state.get_added_validators(next_epoch)` to equal `terminal.added_validators()` and `checkpoint_state.get_removed_validators()` to equal `terminal.removed_validators()` (order-normalized), rejecting mismatches with a new `CheckpointVerificationError::CheckpointTransitionQueueMismatch`.
- Document the inherent residual: pending additions for epochs beyond `next_epoch` (possible when validator warm-up spans more than one epoch) are committed by finalized headers after the terminal and cannot be authenticated by a chain ending at this epoch; they are bound only once the next epoch's boundary header is verified.
- **Test**: add `test_checkpoint_verifier_binds_transition_queues_to_terminal_header` (honest empty-queue checkpoint verifies; a checkpoint carrying a pending `removed_validators` entry while the terminal header commits an empty set is rejected, with hash/membership/position checks still passing). Parameterize `checkpoint_verification_fixture` with the checkpoint's `removed_validators` queue and update its callers.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -658,7 +658,7 @@ impl<
                                 self.db
                                     .get_finalized_header(epoch - 1)
                                     .await
-                                    .map(|h| h.header.digest.0)
+                                    .map(|h| h.header().get_digest().0)
                             } else {
                                 // Future epoch we have not reached: no genesis to serve.
                                 None
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -170,6 +170,17 @@ pub struct RunFlags {
     )]
     pub weak_subjectivity_header_digest: Option<String>,
 
+    /// Import a checkpoint WITHOUT verifying it against a finalized-header chain.
+    ///
+    /// UNSAFE: the imported consensus state is trusted entirely from the supplied
+    /// checkpoint artifact. By default, checkpoint startup requires a checkpoint
+    /// directory containing finalized_headers/ together with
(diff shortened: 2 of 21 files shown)


sebastian-osec:
I think the immediate transition check is correct, but I’m not sure we should accept future `added_validators` buckets as trusted checkpoint contents.

For a checkpoint at epoch `N`, the terminal finalized header authenticates `removed_validators` and `added_validators[N + 1]`. But `ConsensusState.added_validators` is a map, and entries for `N + 2` or later are serialized inside the checkpoint and are not checked here.

Given `VALIDATOR_NUM_WARM_UP_EPOCHS = 2`, this seems reachable for normal deposits: a checkpoint at epoch `N` can carry `added_validators[N + 2]`. A malicious checkpoint could keep `added_validators[N + 1]` correct while choosing future additions, and Step 5 would still pass.

Could we either reject any `added_validators` entries beyond `next_epoch`, or explicitly document that future additions remain trusted checkpoint contents and are outside the fix for #216?

matthias-wright:
Good point, I added a comment in `types/src/checkpoint.rs` to clarify this.

Issue #353, opened 2026-06-01
[Security][tentative] Medium (correctness): Staking Docs Show Zero Withdrawal Address


evonide:
## Context

Summit deposits bind a validator account to withdrawal credentials derived from an Ethereum address. Later withdrawal requests must come from the stored address, so staking examples must not encourage an address that no normal operator controls.

## Claim

A normal operator copies the documented staking request and submits the resulting deposit calldata. The deposit credentials encode address zero, Summit stores address zero as the validator's withdrawal credentials, and later ordinary withdrawal or exit requests from the operator's real address are skipped because the source address does not match the stored credentials. The staking guide's `getDepositSignature` example uses `0x0000000000000000000000000000000000000000`, and the RPC, deposit parsing, and finalizer paths accept and store that address without rejecting it.

## Flow

The trigger reaches Summit through the code paths linked below, is accepted without the missing validation or state update, and is later consumed by the affected consensus, finalizer, sync, or RPC path. The linked code identifies the exact handoff that should preserve the invariant.

## Impact

The operator's stake can become practically unrecoverable through ordinary withdrawal or exit flows because no normal private key controls address zero. This is asset-impacting for the misconfigured validator, but not a remote attacker path unless combined with separate RPC exposure or social engineering.

## Root Cause

The user-facing staking example uses a dangerous placeholder as an executable parameter, and Summit does not reject zero withdrawal addresses at signing, deposit parsing, or account installation.

## Code

- The staking guide shows `getDepositSignature` with the zero address, immediately before telling operators to submit the deposit transaction: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/docs/staking-and-participating.md#L14-L24.
- RPC builds withdrawal credentials from any parsed `Address` without rejecting zero: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/rpc/src/server.rs#L221-L226.
- Summit parses valid-format credentials into the embedded address without a nonzero check: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/utils.rs#L44-L67.
[...]

PR #424, opened 2026-06-18
docs: drop the zero-address reference

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/353.

Changes:
- Drops the zero-address reference from the docs.

Note: this doesn't contain the requested change to `getDepositSignature` on purpose, because the zero-address is still a valid address.


sebastian-osec:
Looks good to me.

Issue #217, opened 2026-05-22
[Security][tentative] High (attackable): Checkpoint Verification Does Not Bind Validator Consensus Keys


evonide:
## Context

Checkpoint verification reconstructs the validator participant set from genesis and finalized-header deltas, then uses those participants to verify historical finalization signatures. Each active validator identity has both a node key and a BLS consensus key, and checkpoint import must preserve that binding so future certificates authorize the same participants.

## Claim

Checkpoint verification uses `(node_public_key, consensus_public_key)` pairs from genesis and finalized-header deltas for signature verification, but the final checkpoint-state consistency check compares node keys only. A checkpoint can therefore pass verification while assigning different BLS consensus keys to the same node identities.

A malicious checkpoint provider with a correctly signed historical checkpoint chain whose terminal header already authenticates the supplied checkpoint bytes can alter BLS consensus keys in the checkpoint state while preserving node identities and active statuses. A passive public distributor cannot swap keys in an honestly signed checkpoint without changing the checkpoint-data hash and failing terminal-header verification.

## Flow

The path is reachable during checkpoint import when verification receives finalized headers signed by the expected historical participants and a terminal header whose `checkpoint_hash` matches the supplied checkpoint bytes, while the embedded account state keeps the same node keys but changes one or more account `consensus_public_key` values. No evidence was found for a public RPC/P2P actor or passive checkpoint mirror changing an existing honest checkpoint and keeping verification passing.

## Impact

A node bootstrapping from that checkpoint installs altered consensus keys into its local state. Future epoch participants are then derived from account keys that were not bound to the verified historical participant set, corrupting validator-set authorization after import.

## Root Cause

The verifier uses full node/BLS pairs for finalized-header signature checks, then reduces the state consistency check to node-key membership instead of comparing the account's consensus key to the accumulated BLS key.

## Code

- The verifier builds participants as node/BLS key pairs from genesis: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L203.
- Finalized-header verification uses the full pair set to construct the multisig verifier: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L230.
- Header deltas also add full node/BLS pairs: https://github.com/SeismicSystems/summit/blob/ed2c5c8/types/src/checkpoint.rs#L249.
[...]

PR #425, opened 2026-06-18
fix: bind checkpoint consensus keys to the verified terminal header

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/423 (which builds on https://github.com/SeismicSystems/summit/pull/422 and https://github.com/SeismicSystems/summit/pull/402).

Addresses https://github.com/SeismicSystems/summit/issues/217.

Changes:
- Step 3 now accumulates a node-key to BLS-key map and requires each checkpoint account's consensus_public_key to equal the BLS key accumulated for that node from the verified history. Mismatch returns the new CheckpointConsensusKeyMismatch error variant. The reverse membership check uses the same map.
- Add test_checkpoint_verifier_binds_consensus_keys_to_terminal_header covering both directions: the honest case (accounts carry the real, accumulated consensus keys and verify) and a swapped consensus key (rejected with CheckpointConsensusKeyMismatch). The shared checkpoint_verification_fixture gains a tamper_consensus_key parameter that swaps one active account's stored consensus key while leaving its node key, status, and the accumulated set intact.

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -658,7 +658,7 @@ impl<
                                 self.db
                                     .get_finalized_header(epoch - 1)
                                     .await
-                                    .map(|h| h.header.digest.0)
+                                    .map(|h| h.header().get_digest().0)
                             } else {
                                 // Future epoch we have not reached: no genesis to serve.
                                 None
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -170,6 +170,17 @@ pub struct RunFlags {
     )]
     pub weak_subjectivity_header_digest: Option<String>,
 
+    /// Import a checkpoint WITHOUT verifying it against a finalized-header chain.
+    ///
+    /// UNSAFE: the imported consensus state is trusted entirely from the supplied
+    /// checkpoint artifact. By default, checkpoint startup requires a checkpoint
+    /// directory containing finalized_headers/ together with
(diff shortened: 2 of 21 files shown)


sebastian-osec:
I think this is a good partial fix for the current active-validator key-swap case, but I think there is still a bypass for validators that are queued to join.

The new check binds `validator_accounts[node_key].consensus_public_key` to the accumulated BLS key only for validators in the current signing set. A `Joining` validator is not in that signing set yet, so its account key is not checked against the verified `added_validators[next_epoch]` entry.

That matters because epoch activation later loads the account by node key and only changes:

```rust
account.status = ValidatorStatus::Active;
```

It does not overwrite `account.consensus_public_key` from the verified `AddedValidator.consensus_key`.

So a malicious checkpoint can make `added_validators[next_epoch]` match the terminal header, while setting the corresponding joining account’s `consensus_public_key` to a different BLS key. Verification accepts it, and when the validator activates, the attacker-chosen account key becomes active.

I think the verifier should also enforce:

```rust
validator_accounts[added.node_key].consensus_public_key == added.consensus_key
```

for every validator in `added_validators[next_epoch]`.

matthias-wright:
Update(bfc4fc9):
- Bind checkpoint consensus keys to the verified terminal header

sebastian-osec:
This looks good to me now.

Issue #310, opened 2026-06-01
[Security][tentative] Medium (attackable): Checkpoint Finalized Header Epoch Replay


evonide:
## Context

Trustless checkpoint import verifies a chain of finalized headers before accepting checkpoint state. The verifier should ensure each header's decoded epoch is the same epoch certified by its finalization certificate.

## Claim

Checkpoint verification enforces header-chain ordering with `finalized_header.header.epoch == i`, but the certificate verifies `finalized_header.finalization.proposal`. Because the header digest omits `epoch`, raw SSZ can mutate the decoded header epoch while preserving the signed payload digest, and `FinalizedHeader::from_ssz_bytes` only checks payload digest equality.

## Flow

The path is reachable through directory checkpoint import with a supplied finalized-header chain. The remaining validation work is to build a minimal raw finalized-header chain that reuses a certificate under a mutated header epoch and observe verifier acceptance.

## Impact

Checkpoint verification can accept a fabricated finalization history and then apply decoded validator deltas from headers whose epoch was not signed for that slot. This can confuse participant-set reconstruction and weaken the trustless checkpoint boundary.

## Root Cause

The verifier mixes unsigned decoded header epoch for ordering with signed proposal data for certificate verification, without checking `finalization.epoch() == header.epoch`.

## Code

- Header digest omits `epoch`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/header.rs#L44-L86
- `FinalizedHeader::from_ssz_bytes` checks that finalization payload equals header digest, but does not bind finalization epoch to header epoch: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/header.rs#L397-L408
- Checkpoint verification enforces contiguity with decoded `header.epoch`: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/checkpoint.rs#L218-L227
- The verifier checks the certificate after that ordering check and then applies decoded validator deltas: https://github.com/SeismicSystems/summit/blob/ed2c5c844e9ec40884ddc72fc16ece36b7d99a56/types/src/checkpoint.rs#L229-L260

## Related Issues/PRs

The following existing items touch the same trust boundary, adjacent root cause, or likely remediation stack.

- #214
- #215
- #218
- #261
- #286
- #294
- #332

[...]

PR #426, opened 2026-06-18
test: add regression for finalized-header epoch-replay rejection

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/425 (which builds on https://github.com/SeismicSystems/summit/pull/423, https://github.com/SeismicSystems/summit/pull/422, and https://github.com/SeismicSystems/summit/pull/402).

Addresses https://github.com/SeismicSystems/summit/issues/310.

This issue was already solved by https://github.com/SeismicSystems/summit/pull/170 and https://github.com/SeismicSystems/summit/pull/286.

This PR just adds a regression test.

Changes:
- Add regression for finalized-header epoch-replay rejection

### DIFF
--- a/finalizer/src/actor.rs
+++ b/finalizer/src/actor.rs
@@ -658,7 +658,7 @@ impl<
                                 self.db
                                     .get_finalized_header(epoch - 1)
                                     .await
-                                    .map(|h| h.header.digest.0)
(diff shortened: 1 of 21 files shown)


sebastian-osec:
Looks good to me.

Issue #311, opened 2026-06-01
[Security][tentative] Medium (attackable): Checkpoint Joining Accounts Gain Peer Authorization


evonide:
## Context

Checkpoint import verifies finalized headers and then installs decoded consensus state as the node's local recovery state. That state contains validator accounts, and Summit's finalizer later uses account status to publish P2P peer-oracle updates for active and joining validators plus derived observer keys.

## Claim

A malicious checkpoint source supplies terminal-header-authenticated checkpoint bytes containing an extra attacker-controlled `validator_accounts` entry with status `Joining`, while the active signing set still matches the finalized-header chain. The imported node starts with active-only oracle tracking, but at a later epoch transition it tracks active-or-joining accounts and derived observers, authorizing the injected key material on consensus channels before the account ever becomes an active voter. `verify_checkpoint_chain` rejects extra `Active` accounts but does not reject extra `Joining` accounts that were never accumulated from genesis or signed finalized-header `added_validators`. After import, those unchecked joining accounts can be tracked into the peer oracle and their derived observer keys can become accepted P2P identities even though they were not authenticated by the verified committee history.

## Flow

The path requires checkpoint startup with bytes already authenticated by the terminal finalized header, or an even weaker unverified checkpoint startup path covered by other findings. Source review shows the verifier ignores extra non-active accounts, `get_active_or_joining_validators()` includes `Joining`, and epoch-transition oracle tracking uses that selector. A runtime reproduction should confirm the exact first epoch where a checkpoint-injected joining account becomes accepted by Commonware discovery and can enqueue consensus-channel traffic.

## Impact

An importing node can accept attacker-chosen primary and observer peer identities from checkpoint state rather than from finalized validator onboarding. When composed with the already-validated consensus-subchannel admission gap, those identities can consume bounded consensus-channel capacity for the importer without BLS voting power. This affects checkpoint importers rather than forcing already-running honest validators to adopt the peer set.

## Root Cause

Checkpoint verification treats non-active validator accounts as outside the signing-set consistency check, while P2P oracle construction later treats `Joining` accounts as network-authorized identities.

## Code

[...]

PR #427, opened 2026-06-18
test: cover the checkpoint_hash binding

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/426 (which builds on https://github.com/SeismicSystems/summit/pull/425, https://github.com/SeismicSystems/summit/pull/423, https://github.com/SeismicSystems/summit/pull/422, and https://github.com/SeismicSystems/summit/pull/402).

Addresses https://github.com/SeismicSystems/summit/issues/311.

This only adds test coverage, the issue is not reachable.
The decoded `ConsensusState`, including every `validator_accounts` entry, is committed by `checkpoint.data`, which `verify_checkpoint_chain` binds to the terminal finalized header via `checkpoint_hash == sha256(checkpoint.data)`. Appending a Joining account changes the digest, so the tampered checkpoint no longer matches the honest terminal header and is rejected at Step 2.

Changes:
- types/src/checkpoint.rs: test_checkpoint_verifier_rejects_extra_joining_account — reproduces the attack directly. It decodes the honest checkpoint, injects an extra Joining account (leaving the active signing set untouched), re-encodes, and asserts verify_checkpoint_chain rejects it with CheckpointHashMismatch.
- application/src/actor.rs: rejects_block_with_mismatched_checkpoint_hash — the consensus-layer half. It mirrors accepts_ordinary_child_inside_epoch, changing only the block's checkpoint_hash, and asserts handle_verify returns false, isolating the checkpoint_hash check as the sole cause.

### DIFF
--- a/application/src/actor.rs
+++ b/application/src/actor.rs
@@ -2243,4 +2243,69 @@ mod tests {
             "honest block with prev_randao = 0 must be accepted"
         );
     }
+
+    /// handle_verify must reject a block whose `checkpoint_hash` disagrees with
+    /// the verifying node's locally-derived `aux_data.checkpoint_hash`. An honest
+    /// validator only votes for a block whose checkpoint_hash matches the
+    /// checkpoint it computed from its own canonical state, so a finalized
+    /// terminal header's checkpoint_hash provably commits to the canonical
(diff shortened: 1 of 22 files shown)


sebastian-osec:
This test commit looks good to me

Issue #232, opened 2026-05-22
[Security][tentative] Medium (attackable): Secondary Observers Can Fill Consensus Subchannels


evonide:
## Context

Summit's P2P layer tracks validator node keys and observer-derived secondary keys for connectivity, while Simplex consensus separately decides which active validators have voting power. Consensus traffic is delivered through bounded epoch subchannels before downstream membership and signature checks, so non-voting peers must not be able to consume validator-only consensus queue capacity.

## Claim

Summit admits non-voting identities into the live P2P peer set used by consensus channels. Observer-derived node keys are tracked as secondary P2P peers, and joining validators are tracked as primary peers before their BLS keys enter the Simplex committee. Because Commonware accepts tracked peers onto shared consensus channels and the epoch mux drops on full subchannels before Simplex validates sender membership, these non-voting peers can consume live consensus subchannel capacity without BLS voting power.

A joining validator key holder, or a holder of that validator's registered observer child key, can connect as an accepted P2P peer and send traffic routed to current consensus subchannels before the validator has Simplex voting power, consuming bounded mux capacity before membership checks can reject the messages.

## Flow

The path is reachable in live networking when a joining validator exists or observer-derived secondary keys are registered and the attacker controls the corresponding private key material. This is not reachable by arbitrary unauthenticated network clients or by unrelated authenticated keys; the attacker needs accepted primary or secondary peer status and enough routed consensus traffic to fill bounded epoch subchannels before honest validator messages arrive.

## Impact

Accepted non-voting peers can fill bounded consensus subchannels, causing honest validator messages for that epoch to be dropped before Simplex membership/signature checks run. With enough joining or observer identities, this can cause consensus liveness loss without requiring BLS voting power.

## Root Cause

P2P peer authorization and consensus voting authorization are not separated before mux delivery. Summit registers joining and observer identities on channels carrying validator-only consensus traffic, and Commonware's mux applies bounded `try_send` backpressure before Simplex validates membership.

## Code

- Summit registers observer-derived keys as secondary peers during startup and epoch transition: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L241, https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L662.
[...]

PR #428, opened 2026-06-19
fix: gate consensus channels by committee membership before the mux

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/232.

Summit tracks non-voting identities (joining validators and observer-derived keys) as secondary P2P peers so they can connect and sync. The authenticated network does not gate per-channel sends by peer role, and the epoch multiplexer drops on a full subchannel (try_send) before Simplex validates sender membership. A non-voting peer can therefore fill a bounded validator-only consensus subchannel and starve honest validator messages. A single staked deposit yields the joining key plus its observer child keys, amplifying the effect.

Changes:
- `orchestrator/src/committee_filter.rs`: `CommitteeFilteredReceiver` wraps a consensus-channel receiver and drops messages whose sender is not in the target epoch's committee, before they reach the bounded mux subchannel. It is epoch-aware via the public `mux::parse`: messages for entered epochs are gated by that epoch's committee; messages for not-yet-entered (ahead) epochs pass through, preserving the pending-channel backup / hint_finalized catch-up path.
- `orchestrator/src/actor.rs`: maintains a per-epoch committee map updated on `Enter` (before the subchannel is registered) and `Exit`, and wraps the pending, recovered, and resolver receivers with the filter. The broadcaster and backfiller channels stay open so joining validators and observers can still sync.
- `node/Cargo.toml`: chain summit-{finalizer,syncer,orchestrator,application}/prom into the node prom feature (previously only summit-types/prom), so all crate metrics actually emit in prom builds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6097,6 +6097,7 @@ dependencies = [
  "commonware-utils",
  "futures",
  "governor",
+ "metrics",
  "rand 0.8.5",
  "rand_core 0.6.4",
  "summit-syncer",
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -154,6 +154,10 @@ prom = [
   "procfs",
   "libc",
   "summit-types/prom",
+  "summit-finalizer/prom",
+  "summit-syncer/prom",
+  "summit-orchestrator/prom",
+  "summit-application/prom",
 ]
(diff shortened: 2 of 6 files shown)


sebastian-osec:
This fixed the issue. Looks good to me.

Issue #236, opened 2026-05-22
[Security][tentative] Medium (correctness): Checkpoint Finalized Header Is Not Seeded Into Finalizer DB


evonide:
## Context

Summit checkpoint startup lets a node resume from checkpoint artifacts instead of replaying all prior history. The syncer needs the checkpoint block artifacts to continue block synchronization, and the finalizer needs the checkpoint epoch's finalized header as the predecessor for future epoch-boundary headers.

## Claim

Checkpoint startup passes the checkpoint `finalized_header` only to the syncer and never seeds it into finalizer storage. The broken invariant is that a checkpoint-joined finalizer can have no persisted predecessor finalized header, causing boundary aux-data generation to fall back to genesis instead of the checkpoint header.

A node operator starts from checkpoint artifacts that include a `finalized_header`, then the checkpoint-joined proposer reaches an epoch boundary before that header is seeded into finalizer storage; aux-data generation reads an empty finalized-header DB, falls back to genesis, and emits boundary blocks linked to the wrong prior header.

## Flow

The path requires startup from checkpoint artifacts that include `finalized_header`, followed by proposing or building an epoch-boundary block before the finalizer DB has the checkpoint header seeded through another path.

## Impact

A checkpoint-joined node can build future epoch-boundary headers whose `prev_epoch_header_hash` points to genesis rather than the checkpoint epoch header. Because boundary header-link verification is also weak, such headers can be accepted and make post-checkpoint finalized-header chains diverge from continuously running nodes.

## Root Cause

Checkpoint import hands the finalized-header artifact to the syncer but not to finalizer storage, and boundary aux-data generation treats a missing finalized header as a genesis fallback.

## Code

- Engine builds `SyncCheckpoint` from `checkpoint_last_block` and `checkpoint_finalized_header` for the syncer: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L341.
- Finalizer initialization receives consensus state and genesis hash, but no checkpoint finalized-header artifact: https://github.com/SeismicSystems/summit/blob/ed2c5c8/node/src/engine.rs#L150.
- Boundary aux data loads `prev_header_hash` from `get_most_recent_finalized_header`, falling back to genesis when the DB is empty: https://github.com/SeismicSystems/summit/blob/ed2c5c8/finalizer/src/actor.rs#L944.
[...]

PR #435, opened 2026-07-02
test: cover checkpoint SyncCheckpoint startup-replay header seeding

matthias-wright:
Addresses https://github.com/SeismicSystems/summit/issues/236.

Changes:
- Adds `test_checkpoint_join_replays_and_seeds_finalized_header`, covering the `SyncCheckpoint` startup-replay path where a checkpoint join carrying `last_block` + `finalized_header` seeds the checkpoint epoch's finalized header into finalizer storage, via the syncer's pre-consensus replay of the terminal block, before the consensus loop starts.

### DIFF
--- a/node/src/tests/checkpointing/joining.rs
+++ b/node/src/tests/checkpointing/joining.rs
@@ -369,6 +369,214 @@ fn test_node_joins_later_with_checkpoint() {
     });
 }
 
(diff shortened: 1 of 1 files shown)


sebastian-osec:
Looks good to me.

PR #437, opened 2026-07-04
feat(rpc): make pause admin address configurable via CLI flag

### LINKED ISSUE #436: Hardcoded pause admin address in rpc auth
rpc/src/auth.rs still contains a hardcoded pause admin address behind a production TODO. I have a fix prepared to make the pause admin configurable via CLI, preserve the current default for compatibility, and fail fast on misconfiguration at startup instead of on the first emergency pause call.

Zhekinmaksim:
Closes #436.

This removes the hardcoded production pause-admin TODO from `rpc/src/auth.rs` by making the address configurable via `--pause-admin-address`. The flag defaults to the current constant, so existing permissioned deployments keep the same behavior unless they opt into a different admin address.

The address is now parsed once at startup and the node fails fast on misconfiguration instead of surfacing an error on the first emergency pause/unpause call. The configured admin address is threaded from `RunFlags` through RPC startup into `SummitRpcServer`, and the permissioned auth path verifies against that explicit address.

I also updated the permissioned RPC integration tests and the node bin helpers that construct `RunFlags` manually so the workspace continues to compile under `--features permissioned`.

Validation:
- `cargo fmt --check`
- `cargo check -p summit-rpc --features permissioned`
- `cargo test -p summit-rpc --features permissioned`
- `cargo check -p summit --features permissioned`
- `cargo clippy --features permissioned`

Opened against `main`, which is the current default branch. If you would prefer this on another active audit branch, I can retarget.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -109,6 +109,11 @@ pub struct RunFlags {
     #[arg(long, default_value_t = 3030)]
     pub rpc_port: u16,
 
+    /// Address authorized to sign pause/unpause RPC requests
+    #[cfg(feature = "permissioned")]
+    #[arg(long, default_value_t = String::from(summit_rpc::DEFAULT_PAUSE_ADMIN_ADDRESS_HEX))]
+    pub pause_admin_address: String,
+
     /// Number of tokio worker threads (defaults to number of logical CPUs)
     #[arg(long)]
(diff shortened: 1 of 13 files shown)


matthias-wright:
https://github.com/SeismicSystems/summit/issues/436#issuecomment-4889613337

PR #452, opened 2026-08-11
fix(cli): correct --prom-port help text

devorun:
`--prom-port` (default `9090`, paired with `--prom-ip` "Prometheus address") sets the port for the Prometheus metrics endpoint, but its help string was copied from `--port` and reads **"Port Consensus runs on"**.

As a result `summit run --help` labels both `--port` and `--prom-port` identically, describing the metrics port as the consensus port. This corrects the `--prom-port` description to the flag it actually configures. `--port` is left unchanged.

Doc-comment only; no behavior change.

### DIFF
--- a/node/src/args.rs
+++ b/node/src/args.rs
@@ -131,7 +131,7 @@ pub struct RunFlags {
     /// Prometheus address
     #[arg(long, default_value_t = String::from("0.0.0.0"))]
     pub prom_ip: String,
-    /// Port Consensus runs on
+    /// Port the Prometheus metrics endpoint runs on
     #[arg(long, default_value_t = 9090)]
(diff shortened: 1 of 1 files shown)


devorun:
Gentle follow-up on this one. I've rebased it onto current `main`, so it's up to date and conflict-free.

It's a one-line help-text fix: `--prom-port` currently shows "Port Consensus runs on" (copied from `--port`), while it actually sets the Prometheus metrics port (default 9090, paired with `--prom-ip`), so `summit run --help` labels both flags identically. `--port` is left untouched.

@matthias-wright would you mind taking a quick look when you have a moment? Happy to adjust the wording if you'd prefer something else. Thanks!

PR #456, opened 2026-08-18
perf: speedup ci by splitting tests into parallel default/all-features jobs

samlaf:
The test job ran `cargo test` and `cargo test --all-features` back to back: 7.8m + 7.7m, about 16m of wall clock. A two-entry matrix runs them side by side and cuts that to about 8m. Each entry gets its own rust-cache shared-key, so the two jobs no longer overwrite each other's cache. `fail-fast: false` keeps a failure in one config from cancelling the other.

Timings from run 32185252971:
```
  cargo test                 7.8m   (2m34s compile, ~5m10s tests)
  cargo test --all-features  7.7m   (1m52s compile, ~5m50s tests)
```

Compilation is only about 30% of the cost. Almost all the rest is one binary: the `summit` lib test target (94 tests in 307s, 96 tests in 287s). Every other crate finishes in under a second, except `summit-syncer` under all-features (27 tests, 56s).

Both configs earn their place. `--all-features` runs a strict superset of the default test names (613 -> 644, and no test is default-only). But three features change behavior in place instead of adding tests:

  - `bench` swaps `start_building_block` for a variant that takes a height and drops the state root (application/src/actor.rs)
  - `bad-blocks` swaps in `BadBlockEngineClient` (node/src/args.rs)
  - `permissioned` changes the RPC auth path (rpc/src/server.rs)

`bench` is on under `--all-features`, so that run never exercises the production block-building path. The default run is what covers it.

## Possible Follow-ups (not done here)

  - `--all-features` turns on `bench` and `bad-blocks` together. No deployment runs that combination. A named feature set, such as `--features prom,permissioned,e2e`, would test a real configuration instead of a frankenconfig.
  - To get below ~8m, shard the `summit` lib binary with `cargo nextest run --partition count:N/M` over a matrix. The slowest tests are multi-node integration tests. Local timings: 207s for test_checkpoint_verification_fixed_committee, 175s for test_process_time_invalid_new_validator_refund_does_not_merge_with_reused_pubkey_withdrawal, 158s for test_deposit_request_top_up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,16 +58,28 @@ jobs:
       - name: check warnings with all features
         run: RUSTFLAGS="-D warnings" cargo check --all-features
 
+  # The two feature sets are not redundant: `bench` and `bad-blocks` swap in
+  # alternative engine clients, so `--all-features` never exercises the
+  # production block-building path that the default run covers.
   test:
+    name: test (${{ matrix.name }})
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: default features
+            features: ""
+            cache-key: "test-cache-default"
+          - name: all features
+            features: "--all-features"
+            cache-key: "test-cache-all-features"
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


samlaf:
TODO: in order to merge will have to replace the now outdated required CI check "test" and make the 2 new ones requires instead.

PR #460, opened 2026-09-14
fix(checkpoint): bind exiting validators to the accumulated signing set

erkancamli:
### Problem

Step 3 of `verify_checkpoint_chain_with_weak_subjectivity` cross-checks the checkpoint's validator accounts against the signing set accumulated from the verified finalized headers. The two directions disagree on which statuses count.

Forward, an accumulated signer is accepted under either signer status:

```rust
if account.status != ValidatorStatus::Active
    && account.status != ValidatorStatus::SubmittedExitRequest
```

Reverse, only one of them has to be accumulated:

```rust
if account.status == ValidatorStatus::Active && !accumulated.contains_key(key) {
```

`ValidatorStatus::is_current_epoch_signer()` is `Active | SubmittedExitRequest`, and `ConsensusState::get_current_epoch_validators` filters on exactly that predicate. `finalizer/src/actor.rs` uses it to seed the epoch committee and the P2P oracle at startup, so both statuses are committee bearing on the importing node, but only one of them is bound to the signing set the headers actually prove.

The consequence is an extra committee member: a checkpoint carrying one additional `SubmittedExitRequest` account with an attacker chosen node and BLS key passes verification, which shifts signer indices and the quorum threshold in `MultisigScheme` and puts the importing node at odds with its honest peers on certificate verification.

This is the threat model Steps 3 to 6 exist for. Step 2's `checkpoint_hash` binding already covers an attacker who tampers with a checkpoint and pairs it with an honest terminal header (that is what `test_checkpoint_verifier_rejects_extra_joining_account` pins). What it cannot cover, per the comment above Step 4, is "a buggy (or colluding) checkpoint creator" that signs a terminal header over its own tampered state. On that path the reverse membership check is what is left.

### Fix

Make the reverse check test the same predicate the forward check accepts:

```rust
if account.status.is_current_epoch_signer() && !accumulated.contains_key(key) {
```

The error message now names the offending status, since it is no longer always `Active`.

### Test

`test_checkpoint_verifier_rejects_extra_committee_account_under_either_signer_status` builds the checkpoint and its terminal header together through the existing `build_checkpoint_and_header` helper, so the hash binding holds and Step 3 is genuinely the check under test. It runs both signer statuses through the same body and asserts `ValidatorSetMismatch` for each.

Before the fix, `Active` is rejected and `SubmittedExitRequest` is not:

```
verifier must reject a checkpoint carrying an extra SubmittedExitRequest account
that never signed any finalized header, got Ok(())
```

After the fix, both are rejected. `cargo test -p summit-types --lib` is 478 passed, 1 ignored, and `cargo test -p summit --lib tests::checkpointing` is 19 passed. The other workspace crates (application, finalizer, syncer, rpc, orchestrator) are green too.

### Environment note

The full `summit` crate lib suite could not be completed in my sandbox: the harness based tests there are killed by the OOM killer partway through, regardless of `--test-threads`. That is a memory limit on my side, not a failure, and it is unrelated to this change, which touches one function in `summit-types`. I ran the checkpointing subtree of that crate on its own and it passes.


erkancamli:
One follow up I would offer on this, because it goes to the cause rather than the instance.

The bug was two spellings of one predicate drifting apart. The fix points the reverse check at `is_current_epoch_signer()`, but the forward check twenty eight lines above it still spells the same rule out as two `!=` comparisons, and so do `consensus_state/mod.rs:1828` (`current_epoch_active_validator_count`) and `mod.rs:2197` (the decode invariant). That leaves exactly the drift hazard that produced this bug.

I checked every committee membership predicate in the file and its callers and they all agree today:

| site | predicate | role |
| --- | --- | --- |
| `checkpoint.rs:545` | `!= Active && != SubmittedExitRequest` | forward check |
| `checkpoint.rs:573` | `is_current_epoch_signer()` | reverse check, fixed here |
| `mod.rs:1809` | `is_current_epoch_signer()` | the actual committee at import |
| `mod.rs:1828` | `matches!(Active \| SubmittedExitRequest)` | exit budget |
| `mod.rs:2197` | `matches!(Active \| SubmittedExitRequest)` | decode invariant |

`get_active_validators` (`== Active`) and `get_validator_keys` (adds `Joining`) look different but are correct: the first is only used at an epoch boundary, after `apply_committee_transition` has already mapped `SubmittedExitRequest` to `FullPayoutPending`, and the second is the p2p peer set rather than voting power.

So collapsing `:545`, and ideally the two in `mod.rs`, onto `is_current_epoch_signer()` would make the symmetry checkable by grep instead of by reading. Want me to fold that into this PR, or keep this one minimal and do it separately?

On the risk of tightening a check, since that is the obvious objection: I could not construct a false rejection. `SubmittedExitRequest` has exactly one producer, the `Active` arm of `apply_withdrawal_request_deferred`, so it implies the account was `Active` earlier in the same epoch and is therefore in `participants`. And `signing_set` is snapshotted at the top of the final loop iteration, before the terminal header's removals are applied, so a genuinely exiting validator is still in it. I also tested it directly: marking a real genesis signer as `SubmittedExitRequest` is still accepted. `cargo test -p summit --lib tests::checkpointing` is 19 passed, including the dynamic committee and joining flows.

erkancamli:
Added the positive test this was missing, and correcting an overstatement in the description.

**The tightening had no test in the direction that would break production.** This narrows *which* accounts step 3 binds, so the case it must keep accepting is an honest checkpoint taken mid-epoch containing a validator that signed this epoch's headers and then submitted an exit request. `test_checkpoint_verification_dynamic_committee` does a full exit at block 15 of epoch 1, but the checkpoint it verifies comes from a later epoch where the account is already `FullPayoutPending`, so nothing pinned the `SubmittedExitRequest`-in-checkpoint case.

`test_checkpoint_verifier_accepts_exiting_account_in_the_accumulated_signing_set` does. It flips a genesis validator, which is in the accumulated signing set, to `SubmittedExitRequest` and asserts the chain still verifies.

I checked it is a real guard and not a duplicate of the negative test: with the production line reverted to `== ValidatorStatus::Active`, the new positive test still passes while the negative one fails with

```
verifier must reject a checkpoint carrying an extra SubmittedExitRequest account that never
signed any finalized header, got Ok(())
```

so the two cover opposite directions. `summit-types` is 479 passed, `tests::checkpointing` 19 passed, `cargo fmt -p summit-types -- --check` clean.

**Correcting the description.** It says a checkpoint carrying an extra `SubmittedExitRequest` account with an attacker-chosen node and BLS key "passes verification", which reads as an unprivileged forgery. It is not. Step 2 binds the checkpoint bytes to `last_header.checkpoint_hash()`, and that header is bound to a verified BLS aggregate certificate, so producing such a checkpoint requires a quorum of the checkpoint epoch's committee to sign a header committing it. That is the same precondition the existing `Active` branch already assumes, and it is the threat model step 3 exists for, but the description should have said so.

What follows from the injected account is still concrete: it reaches `EpochTransition.validator_keys` at startup (`finalizer/src/actor.rs:424-444`), changing committee size, signer indices and quorum. I traced every producer of `SubmittedExitRequest` and the only transition into it is from `Active` in `process_withdrawal_request`, with the removal applied to `participants` only after `signing_set` is snapshotted for the epoch, so an honest mid-epoch checkpoint's exiting accounts are always in `accumulated` and the tightened check cannot reject one.

PR #462, opened 2026-09-17
fix: fix syncer restart replay and preserve recovery history

matthias-wright:
Builds on https://github.com/SeismicSystems/summit/pull/457.

Changes:
 - Resume delivery from the finalizer’s selected startup state, not persisted acknowledgements.
 - Bound finalized-history pruning by the startup recovery height.
 - Remove redundant delivery state and clarify recovery/checkpoint contracts.


matthias-wright:
Cherry-picked on top of https://github.com/SeismicSystems/summit/pull/457

